PR #67, opened 2014-07-23
stylish-haskell.cabal: allow mtl-2.2

trofi:
Signed-off-by: Sergei Trofimovich <slyfox@gentoo.org>


jaspervdj:
Thanks for sending this, I appreciate it. However, looks like it's been fixed.

PR #78, opened 2015-02-21
Introduced new heading for Installation

thefourtheye:


jaspervdj:
Thanks, makes it more readable!

PR #84, opened 2015-04-09
Fix the link to Emacs/Formatting on the HaskellWiki

mineo:
Apparently the wiki fails to redirect properly, resulting in an error page from mediawiki.


jaspervdj:
Thanks!

PR #94, opened 2015-08-12
Adapt to new haskell-src-exts API.

eigengrau:
IVar’s namespace argument has been moved into IAbs as of
HSE b0527bf.


jaspervdj:
I somehow missed this PR, it has been fixed though. Thanks a lot anyway, I appreciate it!

PR #101, opened 2016-04-16
Move CHANGELOG entry for 0.5.9.0 from README

spwhitton:
Move the entry for 0.5.9.0 to where it should be.  Thanks.


jaspervdj:
Oh wow, thanks! I'm not sure how I missed that.

PR #114, opened 2016-07-02
add arch linux to package managers

siddharthist:
https://www.archlinux.org/packages/community/x86_64/stylish-haskell/


jaspervdj:
Thanks!

PR #115, opened 2016-07-02
add more symbols from UnicodeSyntax

siddharthist:
https://downloads.haskell.org/~ghc/7.2.1/docs/html/users_guide/syntax-extns.html


jaspervdj:
Thanks!

PR #116, opened 2016-07-05
Add newline format option

nightuser:
`writeFile` from Haskell uses native newline format, so on Windows all files end with CRLF after running `stylish-haskell`. This PR allows setting line endings (LF or CRLF) optionally. Default value (`native`) gives previous behavior.

### DIFF
--- a/data/stylish-haskell.yaml
+++ b/data/stylish-haskell.yaml
@@ -144,7 +144,16 @@ steps:
 # to. Different steps take this into account. Default: 80.
 columns: 80
 
-# Newline format for output files
+# By default, line endings are converted according to the OS. You can override
+# preferred format here.
+#
+# - native: Native newline format. CRLF on Windows, LF on other OSes.
+#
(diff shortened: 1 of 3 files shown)


jaspervdj:
Hey @nightuser, thanks a lot!  This PR looks great to me.  Two questions:

1. Would it be possible to document the options other than `native` in `stylish-haskell.yaml`?  This file is used as the default configuration but also as the documentation for the configuration.
2. Do we also need to set line ending mode on `stdout` when we are not dealing with a file?  It's not necessary on Linux but I don't know about Windows.

nightuser:
@jaspervdj, I've added some documentation for the option. Changing `stdout` behavior makes sense: `stylish-haskell File.hs > Other.hs` will have correct newline endings too. It can be useful on GNU/Linux / macOS, but I doubt someone will ever want "\r\n" there.

jaspervdj:
Thanks, looking good now!

PR #120, opened 2016-07-19
Added reference to atom-beautify

JustusAdam:
atom-beautify also supports Haskell via stylish-haskell, hence a reference here would probably be good.


jaspervdj:
Thanks!

PR #123, opened 2016-07-31
Empty list align

phadej:
So I can get something like:

```hs
import           Prelude ()
import           Prelude.Compat hiding
                 (foldMap)

import           Data.List
                 (foldl', intercalate)
```

I'd like to make a follow-up so, padding could be specified as `module_name` so it could be *7* or *17*, depending whether there is a qualified import


phadej:
rebased

jaspervdj:
Looks great, thanks!  I appreciate that you added a test case as well.

PR #124, opened 2016-08-03
Add Vim usage method in README

alx741:
I tried this one and it works just fine. It is a better way to use it in Vim.


jaspervdj:
Thanks!

PR #125, opened 2016-08-03
list_padding: module_name

phadej:
The tests show the functionality:

makes it possible not to hardcode `list_padding` to constant value, which is handy with `file` and `group` `align` - variants:

```hs
module Foo where

import Prelude ()
import Prelude.Compat hiding
       (foldMap)

import Data.List
       (foldl', intercalate)
```


jaspervdj:
Thanks!

PR #133, opened 2016-10-11
Add pinned modules feature

phadej:
example, you can _pin_ prelude modules to be on top:

``` hs
import Prelude.Compat hiding (foldMap)
import Prelude ()
import Data.List      (foldl', intercalate, intersperse)
```

### DIFF
--- a/data/stylish-haskell.yaml
+++ b/data/stylish-haskell.yaml
@@ -126,6 +126,14 @@ steps:
       # Default: true
       separate_lists: true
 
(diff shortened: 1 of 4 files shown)


cblp:
Just separate top group from others with extra newline. I disapprove.

phadej:
Well, in my case I'd like `Prelude` to be always on top and then my private `Futurice.Prelude`

even they could be in their own two line group, they'll be in wrong order.

PR #137, opened 2016-11-05
Add a guard for maximum error case

### LINKED ISSUE #136: stylish-haskell breaks at unit record
I got an error.

```bash
$ stylish-haskell test.hs 
stylish-haskell: Prelude.maximum: empty list
$ cat test.hs 
module Main where

data TestData = TestData {} deriving (Show)

main :: IO ()
main = do
  print (TestData {})
  print TestData
$ runghc test.hs
TestData
TestData
```

This is valid haskell code.

mizunashi-mana:
resolve #136


jaspervdj:
Thanks!

PR #165, opened 2017-04-21
Deduplicate import specs

neongreen:
Say if I should add more comments or change something.

(I decided not to make it configurable because after some thinking I couldn't find a reason to do it.)


neongreen:
For examples of deduplication, see tests: https://github.com/jaspervdj/stylish-haskell/pull/165/files#diff-16118401d99390adaef00858a0ed9219R537

jaspervdj:
👍 This is very, very clean code.

PR #167, opened 2017-04-30
Use platform specific configuration paths

jtojnar:


neongreen:
I feel that even if this is released as stylish-haskell-0.9, many users will forget to look at the changelog and their code will be silently formatted differently because the `~/` config will no longer apply. Perhaps a better option is to make xdg a higher priority choice but not remove `~/` entirely?

neongreen:
Also, it'd be nice to say where the xdg config will _usually_ be placed (`~/.config/stylish-haskell/config.yaml`, right?) to save the reader a bit of time and prevent confusion that could arise (it's not clear whether the config will be in `~/.config/stylish-haskell.yaml`, `~/.config/stylish-haskell/stylish-haskell.yaml`, or `~/.config/stylish-haskell/config.yaml`). Also, currently the readme file says `stylish-haskell.yaml` but the code says `config.yaml`.

jtojnar:
I agree that keeping the `$HOME` config is reasonable since the configuration files are merged.

Regarding the location of platform config, I hesitated to mention it in the readme since it is quite verbose and I were not sure how to fit it in the list. Hopefully, this is okay.

PR #168, opened 2017-05-04
Add `imports:space_surround` option

tallen-imvu:
eg:
import Data.Foo ( foo )


tallen-imvu:
redoing this from a branch

PR #169, opened 2017-05-05
Space surround

larver-imvu:
eg:

    import Data.Foo ( foo )

instead of 

    import Data.Foo (foo)


neongreen:
Looks good to me. @jaspervdj?

jaspervdj:
Thanks!

Issue #35, opened 2013-04-07
Break overly-long lines in type constructor list in import list


hseg:
Example:

``` Haskell
import Language.Haskell.Syntax (HsExportSpec(HsEVar, HsEAbs, HsEThingAll, HsEThingWith, HsEModuleContents)
```

should be converted to

``` Haskell
import Language.Haskell.Syntax (HsExportSpec(HsEVar, HsEAbs, HsEThingAll,
                                             HsEThingWith, HsEModuleContents)
```

(Note that I'm starting a new line after the last constructor which is still within 80 chars
from the end of the line.)

ababkin:
Just wanted to bump this since It's 4 years old and is still a very relevant wart. Can this be kindly prioritized? Thanks

dfithian:
👍

PR #176, opened 2017-07-08
Wrap a constructor list to avoid lines too long to fit the screen

ababkin:
addresses https://github.com/jaspervdj/stylish-haskell/issues/35

### DIFF
--- a/lib/Language/Haskell/Stylish/Step/Imports.hs
+++ b/lib/Language/Haskell/Stylish/Step/Imports.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternGuards     #-}
 {-# LANGUAGE RecordWildCards   #-}
 --------------------------------------------------------------------------------
 module Language.Haskell.Stylish.Step.Imports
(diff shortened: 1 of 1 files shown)


ababkin:
A test is failing I see, tried fixing it to no avail. Can I get some help with this?

jaspervdj:
The test case expects:

```haskell
import Data.Identity
    (Identity (Identity, runIdentity))
```

as output, but with your changes we are getting the output:

```haskell
import Data.Identity
    ( Identity (Identity,
                                runIdentity)
    )
```

lib/Language/Haskell/Stylish/Step/Imports.hs
```diff
@@ -372,7 +379,7 @@ prettyImport columns Options{..} padQualified padName longest imp
     mapSpecs f = case importSpecs of
         Nothing -> []     -- Import everything
         Just [] -> ["()"] -- Instance only imports
-        Just is -> f $ map (prettyImportSpec separateLists) is
+        Just is -> f $ map (prettyImportSpec columns separateLists (afterAliasBaseLength + 2)) is
```
jaspervdj: You should use `listPadding'` here rather than `(afterAliasBaseLength + 2)`.

jaspervdj:
I think this code will break when the constructors are not the first thing in the import list.  A better way to do this would be to convert the result type of `prettyImportSpec` to `[String]` so we can also feed it into the `wrap` and `wrapRest` utility functions, that way you wouldn't need to do the computation yourself.

PR #179, opened 2017-08-18
Remove shebang lines

vaibhavsagar:
`nix-shell` has the ability to be used as an interpreter since [version 1.9](https://nixos.org/releases/nix/nix-1.9/manual/#ssec-relnotes-1.9), and the release notes provide an example using Haskell. GHC supports this without any issues, but stylish-haskell only strips the first shebang line and then fails because it cannot parse the file. This provides the ability to strip up to the first two shebang lines.

### DIFF
--- a/lib/Language/Haskell/Stylish/Parse.hs
+++ b/lib/Language/Haskell/Stylish/Parse.hs
@@ -44,11 +44,11 @@ unCpp = unlines . go False . lines
 
 
 --------------------------------------------------------------------------------
--- | Remove shebang from the first line
+-- | Remove shebang lines
 unShebang :: String -> String
-unShebang str
-    | "#!" `isPrefixOf` str = unlines $ ("" :) $ drop 1 $ lines str
-    | otherwise             = str
+unShebang str =
+    let (shebangs, other) = break (not . ("#!" `isPrefixOf`)) (lines str) in
+    unlines $ map (const "") shebangs ++ other
 
 
 --------------------------------------------------------------------------------
--- a/tests/Language/Haskell/Stylish/Parse/Tests.hs
+++ b/tests/Language/Haskell/Stylish/Parse/Tests.hs
@@ -22,6 +22,7 @@ tests = testGroup "Language.Haskell.Stylish.Parse"
     , testCase "Haskell2010 extension"       testHaskell2010
     , testCase "Shebang"                     testShebang
     , testCase "ShebangExt"                  testShebangExt
+    , testCase "ShebangDouble"               testShebangDouble
     , testCase "GADTs extension"             testGADTs
     , testCase "KindSignatures extension"    testKindSignatures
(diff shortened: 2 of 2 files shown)


phunehehe:
Nice! But why limit it to 2 lines? How about "remove shebang until the first line that is not a shebang"?

vaibhavsagar:
GHC doesn't seem to work with shebangs of more than two lines, but the limit is configurable and can be removed if needed.

phunehehe:
I just tried with 3 shebang lines. `runghc` seems to handle that just fine.

vaibhavsagar:
Noted, will remove limit.

lib/Language/Haskell/Stylish/Parse.hs
```diff
@@ -44,13 +44,12 @@ unCpp = unlines . go False . lines
 
 
 --------------------------------------------------------------------------------
--- | Remove shebang from the first line
+-- | Remove shebang lines
 unShebang :: String -> String
 unShebang str
-    | "#!" `isPrefixOf` str = unlines $ ("" :) $ drop 1 $ lines str
+    | "#!" `isPrefixOf` str = '\n':(unShebang $ unlines $ drop 1 $ lines str)
     | otherwise             = str
```
jaspervdj: I think this doesn't need to be recursive -- now it keeps using `lines` and `unlines` which seems a bit wasteful.  We can simply use `break` to find the first non-shebang and then replace the lines before that with the empty string:

```
unshebang :: String -> String
unshebang str =
    let (shebangs, other) = break (not . ("#!" `isPrefixOf`)) (lines str) in
    unlines $ map (const "") shebangs ++ other
```

jaspervdj:
Thanks, this is useful!  I left one small comment.

lib/Language/Haskell/Stylish/Parse.hs
vaibhavsagar: Sure, updated.

lib/Language/Haskell/Stylish/Parse.hs
```diff
@@ -44,11 +44,11 @@ unCpp = unlines . go False . lines
...
 --------------------------------------------------------------------------------
--- | Remove shebang from the first line
+-- | Remove shebang lines
 unShebang :: String -> String
-unShebang str
-    | "#!" `isPrefixOf` str = unlines $ ("" :) $ drop 1 $ lines str
-    | otherwise             = str
+unShebang str =
+    let (shebangs, other) = break (not . ("#!" `isPrefixOf`)) (lines str) in
+    unlines $ map (const "") shebangs ++ other
```
vaibhavsagar: I just realised this could also be `unlines $ ("" <$ shebangs) ++ other` and I can change it to that if you'd prefer.

vaibhavsagar:
Updated :smile:.

jaspervdj:
Thanks!

PR #182, opened 2017-09-10
Add vim-stylishask vim plugin reference

alx741:
The reason for having another plugin is the annoying problem of the code being replaced with errors from *stylish-haskell* when syntax errors exist in the code.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -115,10 +115,18 @@ Or you can define `formatprg`
 
 and then use `gq`.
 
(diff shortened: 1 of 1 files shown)


jaspervdj:
There were some conflicts so I squash-merged this [here](https://github.com/jaspervdj/stylish-haskell/commit/4b9d0745b3da6ced19fb8f257952f317b6d550f9).  Thanks for the plugin and the PR!

Issue #194, opened 2017-12-26
Provide binary script for grabbing stylish-haskell over the network.


lwm:
Following from https://github.com/jaspervdj/stylish-haskell/issues/192 - would you be OK with maintaining a simple bash script here that would do the same as the HLint Travis script and grabs the latest binary and pass some args to it? It would allow the usual `<get program> | sh -s <pass-args>` pattern to make running `stylish-haskell` pretty easy over the network. We can basically fork the scripts that HLint is using:

  1. https://github.com/ndmitchell/hlint/blob/master/misc/travis.sh
  2. https://raw.githubusercontent.com/ndmitchell/neil/master/misc/travis.sh

With a usage like in https://github.com/ndmitchell/hlint#running-with-continuous-integration:

> curl -sL https://raw.github.com/ndmitchell/hlint/master/misc/travis.sh | sh -s .

I can rig up the scripts if so!

jaspervdj:
Sounds good, I'd happily merge the scripts if you can rig them up?

PR #196, opened 2018-01-02
Add scripts to run with pipe recipe.

lwm:
Closes https://github.com/jaspervdj/stylish-haskell/issues/194.

Haven't tested (paths set for when merged) but think this should do the trick!


jaspervdj:
Thanks!  One small nitpick, could we rename `bin/` to `scripts/`?

lwm:
Done!

PR #197, opened 2018-01-06
Fix pattern to match release names.

lwm:
So, it doesn't work yet ;) 

This fix makes the pattern grab the release names you give (`linux` comes before `x86_64`).

Tested with:

```
$ RELEASES=$(curl --silent https://github.com/jaspervdj/stylish-haskell/releases)
$ URL=https://github.com/$(echo $RELEASES | grep -o '\"[^\"]*-linux-x86_64\.tar\.gz\"' | sed s/\"//g | head -n1)
$ echo $URL
```


lwm:
To be clear, if the release naming convention changes, this script will break 👍

PR #207, opened 2018-05-01
Add new step to squash multiple spaces between some elements

hurryabit:
This step basically undoes the effect of `simple_align`. For instance, a case expression like
```
case <expr> of
    Nothing        -> <expr>
    Just something -> <expr>
```
gets squashed to
```
case <expr> of
    Nothing -> <expr>
    Just something -> <expr>
```
Similarly, multiple spaces in record definitions and top level matches get squashed.

It is enabled via
```
- steps:
  [...]
  - squash: {}
  [...]
```
in `.stylish-haskell.yaml`.

---
This change is [Reviewable](https://reviewable.io/reviews/jaspervdj/stylish-haskell/207)

### DIFF
--- a/data/stylish-haskell.yaml
+++ b/data/stylish-haskell.yaml
@@ -198,6 +198,11 @@ steps:
   # Remove trailing whitespace
   - trailing_whitespace: {}
 
+  # Squash multiple spaces between the left and right hand sides of some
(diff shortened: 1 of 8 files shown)


jaspervdj:
Nice, code looks great!

PR #209, opened 2018-05-02
Repsect page breaks alone in a line

### LINKED ISSUE #208: Do not delete page break characters if they are a line's only character
I make heavy use of page breaks to organize and jump around my code. I know it's not a very fashionable thing to do but I think it's a reasonable feature to ask for.

fakedrake:
Page breaks are an old but effective way of bookmarking a file. They
usually are single character lines. Stylish removes them as whitespace
characters, and they are from the compiler's point of view but not
from an editor's perspective.

Resolves #208


jaspervdj:
Thanks!

PR #212, opened 2018-06-15
command in README is not working in current version

bigsleep:
```
curl -sL https://raw.github.com/jaspervdj/stylish-haskell/master/scripts/latest.sh | sh -s .
```
this command is not working in current version.

Install by stack is easy and maybe no one using this command.

I fixed the script, but maybe just remove from README is another option.


jaspervdj:
@lwm Can you review this?  LGTM

PR #216, opened 2018-07-09
Implemented a new import style "semicolons".

joecrayne:
With this style of formatting the import list, multiple imports
of the same library are indented as a single import stanza.  For,
example,

```
import qualified Data.Map as Map
         ;import Data.Map (Map)
````

A semicolon is prepended so that the compiler treats the indented
line as a top-level declaration.


jaspervdj:
I haven't seen this style before, are there any projects using this?  It looks very interesting.

In order to review this PR, it would need to have at least some tests as well, so this style doesn't get broken in the future.

PR #226, opened 2018-12-06
Update bounds on yaml and aeson

DanBurton:
When I said in https://github.com/jaspervdj/stylish-haskell/pull/225 that I had tested against nightly-2018-12-05, I was apparently mistaken. These constraints also need to be relaxed in order to be compatible with the new `aeson` and `yaml` releases.


brandon-leapyear:
Also #227

PR #234, opened 2019-04-29
Bump `haskell-src-exts` upper bound

fosskers:
When using `1.21.0`, this allows for correct parsing of `DerivingVia` syntax.

Keep in mind that it's up to the person building `stylish-haskell` to ensure that the most up-to-date version of `haskell-src-exts` is being used :)

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-flags: {}
+resolver: lts-13.19
+
 packages:
-- '.'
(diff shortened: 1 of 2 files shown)


NickSeagull:
@jaspervdj is there something I can do in order to help merging this?

jaspervdj:
Thanks for reminding me!  Released as `v0.9.2.2` now.

Issue #210, opened 2018-05-29
Enable extensions implied by other extensions


qrilka:
`stylish-haskell` fails to parse module with `type` in import saying 
```
Please add {-# LANGUAGE ExplicitNamespaces #-} pragma at the top of your module."
```
at the same time there is `{-# LANGUAGE TypeOperators      #-}` in that module so according to https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#explicit-namespaces-in-import-export this extension should be already implied

safareli:
Also even if you specify TypeOperators or even ExplicitNamespaces in package.yaml it doesn't care and still fails.

jaspervdj:
`stylish-haskell` currently does not take implied extensions into account.

This should be fixed using `haskell-src-exts`'s `impliedExts` function from `Language.Haskell.Exts.Extension`.

It's probably a relatively easy task -- happy to review a PR if anyone feels up to this.

basile-henry:
I believe `haskell-src-exts` already adds the implied extensions when parsing. The issue is that the version of `haskell-src-exts` currently used with stackage LTS 13.19 is `1.20.3` and the implied extensions for `TypeOperators` are only in `1.21.0`.

The following parses after upgrading to stackage LTS 14.4 (but not before):

```haskell
{-# LANGUAGE TypeOperators #-}

import           GHC.TypeNats (type (<=))
```

I'll open a PR to upgrade to LTS 14.4 :smile:

PR #237, opened 2019-09-02
Update to Stack LTS 14.4

basile-henry:
In this new snapshot we have among other things a more recent version of `haskell-src-exts` which fixes a few of the parsing issues `stylish-haskell` has.

It definitely solves #210 but might also solve other issues like the ones to do with DerivingVia parsing.


jaspervdj:
Thanks!

Issue #138, opened 2016-11-11
centralisation of extensions/pragmas


mwotton:
I use a few tools like stylish-haskell, hlint etc - all of them need to know what extensions to use. While I could put them all in every file, i generally have a list of extensions i want to use everywhere in the project, and would prefer to centralise them _somewhere_. is there some way of pulling them from .ghci, the cabal file, or stack.yaml? i really don't care which, but duplicating them for every tool is a real pain.

jaspervdj:
Yes, `cabal` would be the best place to store these.  Unfortunately it is not simple to parse cabal files.  That is why I usually put the required extensions at the top of every file -- that should be understood by every tool.

mwotton:
yes, that's an option, but it becomes very unwieldy when you have many files and many extensions.

mwotton:
would you accept a PR that added a yaml file with a list of extensions, or a prelude to be silently prepended to each file?

jaspervdj:
I'd rather not add an additional file, that's just creating another standard.  Would it be possible to parse it from the cabal file, either using the Cabal library, or just a simple custom parser that recognizes extension names.

keremispirli:
With the introduction of [hpack](https://github.com/sol/hpack), such extensions can be listed under `default-extensions` in  `package.yaml` file. 

From the ReadMe file:

>  Both `cabal2nix` and `stack` support `package.yaml` natively.

But unfortunately I could not get it working with `stylish-haskell` as of 0.9.0.2.

I use VS Code with [stylish-haskell](https://marketplace.visualstudio.com/items?itemName=vigoo.stylish-haskell) addon. I had to disable it since migrating to `hpack` because I get errors like:
`Language.Haskell.Stylish.Parse.parseModule: could not parse (...)\src\API.hs: ParseFailed (SrcLoc "<unknown>.hs" 34 1) "TypeOperators language extension is not enabled. Please add {-# LANGUAGE TypeOperators #-} pragma at the top of your module."`

Is this because stylish-haskell doesn't support it yet or the add-on fails to pass something to stylish-haskell?

matt-snider:
@keremispirli  I also use hpack and the VS Code stylish-haskell addon. I get the exact same errors. Did you find any way of getting stylish-haskell to consider the package.yaml file?

keremispirli:
[...]

PR #241, opened 2019-10-24
Work on resolving #138

geo2a:
I suggest using Cabal-the-library to find and parse the `.cabal` file in the project root and extract the contents of all `default-extensions` fields. These extensions are appended to ones specified in `.stylish-haskell.yaml` with duplicates removed.

In case of formatting a file outside of a project, i.e. when there is no related `.cabal` file in any of the parent directories, `stylish-haskell` will fallback to the extensions specified in `.stylish-haskell.yaml` and in the file itself.  

I use `stylish-haskell` via `haskell-mode` in Emacs with the default configuration and the fix works fine for me, but perhaps I am too biased towards my workflow?

@jaspervdj could you please suggest an approach I could use to write tests for this kind of functionality?


jaspervdj:
Hi @geo2a, this is awesome!  I will give it a test drive but it looks like it should work on most setups.

I may refactor it a little bit to put the Cabal-related parts in a separate module, but other than that it looks pretty good.

For testing, there are two options that come to mind:

1. Create a directory structure in the `tests/` folder, check it in and test on that directly.
2. Within the test, use `System.Directory.getTemporaryDirectory` and create a directory `stylish-haskell-RANDOM_NUMBER`, then create a directory hierarchy in there with a cabal file, and use that for testing.

I prefer (2) since it's a bit more flexible, and it keeps the tests "all Haskell".

geo2a:
Thanks @jaspervdj! I'll go with the second approach for testing and update the PR sometime during the weekend.

tests/Language/Haskell/Stylish/Config/Tests.hs
```diff
@@ -0,0 +1,144 @@
...
+-- | Perform an action inside a temporary directory tree and purge the tree afterwords
+withTestDirTree :: IO a -> IO a
+withTestDirTree action = do
+  current <- getCurrentDirectory
+  temp <- createTempDirectory "stylish_haskell"
+  setCurrentDirectory temp
+  result <- action
+  setCurrentDirectory current
+  removeDirectoryRecursive temp
+  pure result
```
jaspervdj: Nitpick: you can use `bracket` to clean this up even in case there's exceptions:

```haskell
withTestDirTree :: IO a -> IO a
withTestDirTree action = bracket
    ((,) <$> getCurrentDirectory <*> createTempDirectory "stylish_haskell")
    (\(current, temp) -> do
        setCurrentDirectory current
        removeDirectoryRecursive temp)
    (\(_, temp) -> setCurrentDirectory temp >> action)
```

geo2a: Indeed! I have refactored `withTestDirTree` to use `bracket` and Applicative notation.

jaspervdj:
Thanks a lot!

PR #242, opened 2019-10-28
Add a cabal option to control cabal parsing

jaspervdj:


jaspervdj:
Hey @geo2a, I've moved the Cabal reading code into a separate module because the `Config` module was becoming rather large, and I've also added a `cabal` field in the configuration so users can disable `.cabal` file parsing if they want to.  Can you give this a quick review at some point?

lib/Language/Haskell/Stylish/Config/Cabal.hs
```diff
@@ -0,0 +1,93 @@
...
+--------------------------------------------------------------------------------
+findLanguageExtensions :: Verbose -> IO [Language.KnownExtension]
+findLanguageExtensions verbose =
+    findCabalFile verbose >>=
+    maybe (pure []) (readDefaultLanguageExtensions verbose)
+
+
+--------------------------------------------------------------------------------
+-- | Find the closest .cabal file, possibly going up the directory structure.
+--   It's essential that
```
geo2a: Oops, sorry, "It's essential that" is probably a part of a thought that didn't fully make it into the docstring from my head. I think it's better to just remove it before merging.

stylish-haskell.cabal
```diff
@@ -128,8 +132,8 @@ Test-suite stylish-haskell-tests
     aeson            >= 0.6    && < 1.5,
     base             >= 4.8    && < 5,
     bytestring       >= 0.9    && < 0.11,
+    Cabal            >= 2.4    && < 2.5,
```
geo2a: I am very hazy on the concept of version bounds, could you quickly explain the motivation for [2.4, 2.5) interval?

jaspervdj: So what we're doing is trying to specify bounds to prevent tools from trying to build stylish-haskell with the wrong version of Cabal.  You and both tested this with 2.4, so that's fine.  Any further release in the 2.4.X range _must_ be compatible with 2.4.

However, Cabal 2.5 is not out yet, and since that bumps a major version, it may break the part of the API we're using. Therefore, I prefer to error on the side of caution here.  If 2.5 comes out and it's compatible, it's a quick fix anyway.

geo2a: Thanks for clarifying on this @jaspervdj !

PR #246, opened 2019-11-14
Allow case insensitive language pragmas 🔤

### LINKED ISSUE #245: Allow lowecase on language pragmas
Hi Jasper!

`stylish-haskell` is great, thanks for creating it! 🎉 

Some people (I'm looking at you @serras) prefer to write language pragmas en lowercase, maybe an option to preserve case could be added to:
```
  - language_pragmas:
      preserve_case: true
```

If this is easy enough I can provide a PR! 😄 

Of course this is not critical or anything, I just want to have a stronger argument against convincing Alejandro to use `stylish-haskell` everywhere 😂

kutyel:
Fixes #245 

---
@serras 👀


lib/Language/Haskell/Stylish/Config.hs
```diff
@@ -260,5 +259,7 @@ parseTrailingWhitespace _ _ = return TrailingWhitespace.step
 
 --------------------------------------------------------------------------------
 parseUnicodeSyntax :: Config -> A.Object -> A.Parser Step
-parseUnicodeSyntax _ o = UnicodeSyntax.step
-    <$> o A..:? "add_language_pragma" A..!= True
+parseUnicodeSyntax c o = UnicodeSyntax.step
+    <$> o A..:? "add_language_pragma"    A..!= True
+    <*> o A..:? "allow_lowercase_pragma" A..!= allowLowercase
+    where allowLowercase = elem "allow_lowercase" $ configLanguageExtensions c
```
kutyel: I have serious doubts about this part, @jaspervdj is there a better way to access config flags from other parts?

lib/Language/Haskell/Stylish/Config.hs
```diff
@@ -237,8 +235,9 @@ parseLanguagePragmas :: Config -> A.Object -> A.Parser Step
 parseLanguagePragmas config o = LanguagePragmas.step
     <$> pure (configColumns config)
     <*> (o A..:? "style" >>= parseEnum styles LanguagePragmas.Vertical)
-    <*> o A..:? "align" A..!= True
+    <*> o A..:? "align"            A..!= True
     <*> o A..:? "remove_redundant" A..!= True
+    <*> o A..:? "language_prefix"  A..!= "LANGUAGE"
```
jaspervdj: Can we factor out this line into a separate function that calls `fail` when the prefix is invalid (i.e. not "language")?

lib/Language/Haskell/Stylish/Config.hs
```diff
@@ -262,3 +261,4 @@ parseTrailingWhitespace _ _ = return TrailingWhitespace.step
 parseUnicodeSyntax :: Config -> A.Object -> A.Parser Step
 parseUnicodeSyntax _ o = UnicodeSyntax.step
     <$> o A..:? "add_language_pragma" A..!= True
+    <*> o A..:? "language_prefix"     A..!= "LANGUAGE"
```
jaspervdj: We can then reuse that function here

lib/Language/Haskell/Stylish/Step/LanguagePragmas.hs
```diff
@@ -87,10 +83,14 @@ truncateComma xs
...
-prettyPragmas cols _       _     Compact     = compactPragmas cols
-prettyPragmas cols _       align CompactLine = compactLinePragmas cols align
+prettyPragmas :: String -> Int -> Int -> Bool -> Style -> [String] -> Lines
+prettyPragmas lp _    longest align Vertical    = verticalPragmas lp longest align
+prettyPragmas lp cols _       _     Compact     = compactPragmas lp cols
+prettyPragmas lp cols _       align CompactLine = compactLinePragmas lp cols align
+
+--------------------------------------------------------------------------------
+makeLang :: String -> String
+makeLang x = if fmap toLower x == "language" then x else "LANGUAGE"
```
jaspervdj: And then this shouldn't be necessary anymore, which I think is nice because erroring out is usually better than implicitly changing what the user wanted

kutyel:
@jaspervdj I hope this looks fine! 🙌

jaspervdj:
Thanks again!

PR #249, opened 2019-11-21
New Logo

rakestto:
I've created a new logo for this prettifier! It's a Lambda with others Lambdas within it and a drop of ink. I wish you like it


README.markdown
```diff
@@ -1,29 +1,28 @@
-stylish-haskell
-===============
+## stylish-haskell
+
+<img src="./assets/Logo/SVG/RoundedLogo.svg" width="100px">
```
kutyel: Very cool logo indeed 👌 hope that @jaspervdj likes it! ❤️ 

stylish-haskell

rakestto: Thank you Flavio!!

jaspervdj:
I quite like these!  We should probably pick a single one though.  Do you have a preference yourself @rakestto?

rakestto:
Thank you @jaspervdj I'm glad you like it. The rounded logo included in Readme file is my favourite for web or favicon. The other two logos, whitout rounded background are the best option for extensions like vscode-icons, I've made for that. My favourite between both options is the pink one. If you want to choose just one, I think the pink version without round background is the most dynamic option.

jaspervdj:
Right -- it makes sense that they all have different uses.  I'll merge this as-is; I don't think I'll use the white one but let's keep it around anyway.  Thanks again for this awesome work!!

PR #254, opened 2020-01-09
Allow setting "columns: null" to disable all wrapping

chris-martin:
In some projects I permit arbitrarily long lines, and so I do not want the `columns` setting.

This PR changes the type of the `configColumns` field from `Int` to `Maybe Int`, and allows you to write `columns: null` in `.stylish-haskell.yaml` to disable all line wrapping features.

In addition to the modification to the `Config` type, there are also many functions with an argument type that changes from `Int` to `Maybe Int`, so this would be a significant breaking change to the library.

A workaround without this change would be to simply set some very large number, e.g. `columns: 9223372036854775807`, but that seems rather inelegant.

I did not fully understand what I was doing in the `Util` module, so that part in particular may be worth reviewing.

### DIFF
--- a/data/stylish-haskell.yaml
+++ b/data/stylish-haskell.yaml
@@ -223,7 +223,11 @@ steps:
   # - squash: {}
 
 # A common setting is the number of columns (parts of) code will be wrapped
-# to. Different steps take this into account. Default: 80.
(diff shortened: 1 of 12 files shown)


jaspervdj:
This looks good, thanks a lot!

Issue #39, opened 2013-07-21
Comments should not be stripped from 'import' statements and language pragmas


thoughtpolice:
Consider:

``` haskell
module Foo where
import Data.Monoid -- from base
import Data.List   -- from base
```

And then:

```
$ stylish-haskell /tmp/Foo.hs
module Foo where
import           Data.List
import           Data.Monoid
$
```

The comments shouldn't be stripped from the end, ideally. At the very least this is inconsistent with the default behavior which will preserve comments in most other positions.

siddharthist:
Note that this also happens on `LANGUAGE` pragmas. 
```haskell
{-# LANGUAGE ScopedTypeVariables #-} -- test
```
becomes
```haskell
{-# LANGUAGE ScopedTypeVariables #-}
```

andrewufrank:
i observed the same "disappearing" comments on language pragmas and import. please fix soon - is required to document! (and putting them on separate lines makes things less `stylish`
thank you for a useful tool!

neongreen:
It also happens with comments in the middle of `import` statements, e.g. https://github.com/jaspervdj/stylish-haskell/issues/151:

```haskell
{-# OPTIONS_GHC -F -pgmF htfpp #-}
module Main where

import Test.Framework
import Test.Framework.BlackBoxTest
import {-@ HTF_TESTS @-} MyPkg.A
import {-@ HTF_TESTS @-} MyPkg.B

main = htfMain htf_importedTests
```

andrewufrank:
any fix in sight ? - it makes it impossible to use stylish haskell with HTF testing (and HTF testing is in many projects crucial). i am sorry that i had to de-install stylish...

ChShersh:
@andrewufrank De-installing `stylish` is too radical solution. It would be better if instead you fix `stylish-haskell` ;)
Though I'm also find it rather unconvenient when comments are stripped out. @AnneTheAgile how is it going in 2017 since 2014? I saw you would like to contribute such a nice small thing :)

eflister:
+1 just ran into this, please fix!  note hindent also has bad comment problems :(  https://github.com/commercialhaskell/hindent/issues/370

EncodePanda:
Anyone here willing to write PR for this? I'm open for some pairprogramming session via hangouts to get you started.

PR #256, opened 2020-01-17
Support for records formatting

lukasz-golebiewski:
(Optionally) turns this:
```
module Herp where

data Foo a = Foo { a :: Int, a2 :: String } | Bar { b :: a } 
```
into
```
module Herp where

data Foo a = Foo 
  { a :: Int
  , a2 :: String 
  } 
  | Bar
  { b :: a
  }
```

### DIFF
--- a/data/stylish-haskell.yaml
+++ b/data/stylish-haskell.yaml
@@ -16,8 +16,7 @@ steps:
   #     add_language_pragma: true
 
   # Format record definitions
-  - records:
-      indent: 4
+  - records: {}
 
   # Align the right hand side of some elements.  This is quite conservative
   # and only applies to statements where each element occupies a single
@@ -225,6 +224,9 @@ steps:
   # elements into single spaces. Basically, this undoes the effect of
   # simple_align but is a bit less conservative.
   # - squash: {}
+
+# A common indentation setting. Different steps take this into account.
+indent: 4
 
 # A common setting is the number of columns (parts of) code will be wrapped
 # to. Different steps take this into account.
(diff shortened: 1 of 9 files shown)


data/stylish-haskell.yaml
```diff
@@ -15,6 +15,10 @@ steps:
   #     # true.
   #     add_language_pragma: true
 
+  # Format record definitions
+  - records:
+      indent: 4
```
jaspervdj: I think I would move the `indent` setting to the top-level, just like `columns`, so that other steps may possibly use this value?

lukasz-golebiewski: Good idea :-) Done
Thanks for the review!

EncodePanda:
Hi @jaspervdj, is there anything else you would like us to do in this PR or is it ready to be merged?

jaspervdj:
Hey @EncodePanda, the code looks good to me now; I just need to find some time to manually play with this, try it on a few larger codebases and then it's good to go in -- I promise I'll work on getting that done before the end of the week.

lukasz-golebiewski:
Thanks for the review and the approve @jaspervdj !
We've noticed one "slight inconvenience" which is the removal of comments if any existed somewhere inside the data definition. We're going to spend some time and try to make sure they are preserved

jaspervdj:
Okay, great that you caught that!  Another option you could consider is not formatting the data definition at all if there are any comments inside.

lukasz-golebiewski:
Good idea @jaspervdj! Looks like for now we can do what you suggested only for Haddock style comments, due to how https://hackage.haskell.org/package/haskell-src-exts-1.23.0/docs/Language-Haskell-Exts-Comments.html#v:associateHaddock works

Looks like the problem with comments being removed is more general (see https://github.com/jaspervdj/stylish-haskell/issues/39) and probably needs to be solved in a separate PR

EncodePanda:
We are almost there :) For now we will not format records that have comments in them but we have already a WIP that most likely will fix that (will arrive as a separate PR)

PR #257, opened 2020-01-23
Fix records with comments

EncodePanda:
After this change records with comments will be formatted and comments will *not* get deleted


tests/Language/Haskell/Stylish/Step/Data/Tests.hs
```diff
@@ -298,71 +319,73 @@ case15 = expected @=? testStep (step 2) input
...
        ]
     expected = unlines
        [ "module Herp where"
        , ""
        , "data Foo a = Foo"
-       , "  { a :: a,"
-       , "-- ^ comment"
-       , "   a2 :: String"
+       , "  { a :: a"
+       , "   -- comment"
```
lukasz-golebiewski: This comment is a bit off to the right, can you move it 1 space to the left?

EncodePanda: 👍

Issue #200, opened 2018-01-29
Option to ignore certain directory


denibertovic:
Is there an option in the config file that I can turn on for the formatter to ignore an entire directory and/or project? 

My use case is with using vim-stylish-haskell to run the formatter on each buffer save. That works out for all of my projects except a huge legacy one that I would want to avoid reformatting.

Issue #162, opened 2017-04-15
Add a recursive mode


neongreen:
This will be useful for things like #107.

peterbecich:
It has this
```
-r,--recursive           Recursive file search
```
but it doesn't work as I expect. Maybe I am using it incorrectly.

PR #264, opened 2020-01-30
Recursing directories to find Haskell source files.

mmzx:
This could be one solution to #162 .

There is an added *-r* command-line switch that accepts a folder as its input to search for Haskell source files recursively. In a CI script the already suggested solution in #162 is feasible, however in case _stylish-haskell_ can solve it itself the CI script would become easier.
 
There are pending questions:
- whether you are happy with logging, that is only available in when verbose mode is enabled.
- Could we combine this implementation right away together with #200 ?
- Should the application default to Haskell file extension(s) by default implicitly or should be there another optional argument to specify file extension(s)?

### DIFF
--- a/lib/Language/Haskell/Stylish.hs
+++ b/lib/Language/Haskell/Stylish.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE LambdaCase #-}
 --------------------------------------------------------------------------------
 module Language.Haskell.Stylish
     ( -- * Run
@@ -114,8 +115,7 @@ format maybeConfigPath maybeFilePath contents = do
 findFiles :: Bool -> Maybe FilePath -> IO [FilePath]
 findFiles _ Nothing    = return []
 findFiles v (Just dir) = do
-  existsDir <- doesDirectoryExist dir
-  case existsDir of
+  doesDirectoryExist dir >>= \case
     True  -> findFilesRecursive dir >>=
       return . filter (\x -> takeExtension x == ".hs")
     False -> do
@@ -131,8 +131,7 @@ findFiles v (Just dir) = do
       ps <- listDirectory topdir >>=
         mapM (\x -> do
                  let path = topdir </> x
-                 existsDir <- doesDirectoryExist path
-                 case existsDir of
+                 doesDirectoryExist path >>= \case
                    True  -> go path
                    False -> return [path])
       return $ concat ps
--- a/lib/Language/Haskell/Stylish.hs
+++ b/lib/Language/Haskell/Stylish.hs
@@ -11,7 +11,7 @@ module Language.Haskell.Stylish
     , trailingWhitespace
     , unicodeSyntax
       -- ** Helpers
-    , findFiles
+    , findHaskellFiles
     , stepName
       -- * Config
     , module Language.Haskell.Stylish.Config
@@ -113,14 +113,14 @@ format maybeConfigPath maybeFilePath contents = do
 
 --------------------------------------------------------------------------------
 -- | Searches Haskell source files in any given folder recursively.
-findFiles :: Bool -> [FilePath] -> IO [FilePath]
-findFiles v fs = mapM (findFilesR v . Just) fs >>= return . concat
+findHaskellFiles :: Bool -> [FilePath] -> IO [FilePath]
+findHaskellFiles v fs = mapM (findFilesR v) fs >>= return . concat
 
 
 --------------------------------------------------------------------------------
-findFilesR :: Bool -> Maybe FilePath -> IO [FilePath]
-findFilesR _ Nothing    = return []
-findFilesR v (Just dir) = do
+findFilesR :: Bool -> FilePath -> IO [FilePath]
+findFilesR _ []   = return []
+findFilesR v dir = do
   doesFileExist dir >>= \case
     True -> return [dir]
     _    -> doesDirectoryExist dir >>= \case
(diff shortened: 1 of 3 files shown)


mmzx:
This is a WIP, at least until the pending  questions are answered.

lukasz-golebiewski:
Hi @mmzx ! Thanks for your contribution!
1) Logging only when the `verbose` flag is set sounds right to me
2) Not sure about this. What would be the benefit of doing so over committing both changes separately? 
3) I'd say filtering on `*.hs` is fine for now, we can always add custom extension support later

@jaspervdj @EncodePanda WDYT?

mmzx:
>     2\. Not sure about this. What would be the benefit of doing so over committing both changes separately?

No benefit. It shall be in a separate PR, the occassion was an easy way for me to signal it here. :)

mmzx:
Thanks for all feedback so far!

I have just created a commit that resolves #200, for the time being on a separate [remote branch](https://github.com/mmzx/stylish-haskell/tree/blacklist). Actually there would be an advantage to cover this here: the function which recurses directories is utilized in case the provided input would be a directory. So, the two solutions do depend on each other.

Please let me know how to proceed: a separate PR or moving on here.

lukasz-golebiewski:
I would opt for doing those two separately

mmzx:
So be it: #270 . :)

lukasz-golebiewski:
Could you add a test for this?

mmzx:
That will need a bit of refactoring, because I have to mode the functions to be tested into separate modules; but sure I can do that.

mmzx:
Okay, these commit should indicate my idea. Tomorrow I will update the test suite with working instance(s).

mmzx:
By now I think the code is in a good state.

jaspervdj:
Hey, I think the code looks good but I still think that `-r` should be a boolean that impacts the paths given to `stylish-haskell`.  This is what other command line tools do; for example `ls`:

> ls -R src/ tests/

So basically:

 -  Make `saRecursive` a boolean
 -  If it's set, `mapM` every file in `saFiles` using `findFiles`

mmzx:
This has now a property that recursively it only finds _hs_ files, however explicitly given files are preserved.

lib/Language/Haskell/Stylish.hs
```diff
@@ -103,3 +109,37 @@ format :: Maybe ConfigPath -> Maybe FilePath -> String -> IO (Either String Line
...
+--------------------------------------------------------------------------------
+-- | Searches Haskell source files in any given folder recursively.
+findFiles :: Bool -> [FilePath] -> IO [FilePath]
+findFiles v fs = mapM (findFilesR v . Just) fs >>= return . concat
+
+
+--------------------------------------------------------------------------------
+findFilesR :: Bool -> Maybe FilePath -> IO [FilePath]
+findFilesR _ Nothing    = return []
+findFilesR v (Just dir) = do
```
jaspervdj: Why is this a `Maybe`?  It doesn't seem necessary (or at least it's clear to me why it should be a `Maybe`).

lib/Language/Haskell/Stylish.hs
```diff
@@ -103,3 +109,37 @@ format :: Maybe ConfigPath -> Maybe FilePath -> String -> IO (Either String Line
 format maybeConfigPath maybeFilePath contents = do
   conf <- loadConfig (makeVerbose True) (fmap unConfigPath maybeConfigPath)
   pure $ runSteps (configLanguageExtensions conf) maybeFilePath (configSteps conf) $ lines contents
+
+
+--------------------------------------------------------------------------------
+-- | Searches Haskell source files in any given folder recursively.
+findFiles :: Bool -> [FilePath] -> IO [FilePath]
+findFiles v fs = mapM (findFilesR v . Just) fs >>= return . concat
```
jaspervdj: Let's rename this to `findHaskellFiles`?

src/Main.hs
```diff
@@ -99,14 +105,17 @@ stylishHaskell sa = do
...
             let steps = configSteps conf
             forM_ steps $ \s -> verbose' $ "Enabled " ++ stepName s ++ " step"
             verbose' $ "Extra language extensions: " ++
                 show (configLanguageExtensions conf)
-            mapM_ (file sa conf) files'
+            mapM_ (file sa conf) $ files' filesR
   where
     verbose' = makeVerbose (saVerbose sa)
-    files'   = if null (saFiles sa) then [Nothing] else map Just (saFiles sa)
+    files' x = if null x then [Nothing] else map Just x
```
jaspervdj: We probably want to change this... I think if you would run something like `stylish-haskell -r dir-with-no-haskell-files/`, `stylish-haskell` would end up with no files here and wait for `stdin`, which is not what the user meant.  We should change the code here to use `[Nothing]` if:
1. `-r` is not set (alternatively, we could use `.` in that case)
2. No file paths are explicitly given.

lib/Language/Haskell/Stylish.hs
mmzx: Right. I have overlooked it.

lib/Language/Haskell/Stylish.hs
mmzx: Sure.

src/Main.hs
mmzx: I am puzzling on this one. Please confirm whether my interpretation is correct:
When `-r` is set then `stdin` shall not be considered at all.

I would not involve using `.`.

jaspervdj: Correct: when `-r` is set `stdin` should not be considered.

mmzx:
I believe this would bring the least amount of changes, yet keeps the original behavior.

jaspervdj:
Thanks @mmzx!

mmzx:
Thanks a lot! :)

Issue #262, opened 2020-01-27
Enum formatting


jaspervdj:
`v0.10.0.0` formats the following:

```
data Tag = Title | Text
```

As:

```
data Tag = Title
    | Text
```

Whereas it should really just stay as-is.  I suspect this may be a problem with the new records formatting?  cc @lukasz-golebiewski @EncodePanda

EncodePanda:
Yeah, we saw that the other day. We will fix that ASAP.

PR #265, opened 2020-02-01
Disable formatting of data types without records

maksbotan:
Perhaps the safest way for now is to revert to old behavior for data
types -- i.e. do nothing. Attemmpt reformatting only if at least one
constructor has record fields.

Fixes https://github.com/jaspervdj/stylish-haskell/issues/262

### DIFF
--- a/tests/Language/Haskell/Stylish/Step/Data/Tests.hs
+++ b/tests/Language/Haskell/Stylish/Step/Data/Tests.hs
@@ -395,5 +395,5 @@ case20 = input @=? testStep (step 2) input
     input = unlines
        [ "module Herp where"
        , ""
-       , "data Tag = Title | Text"
+       , "data Tag = Title | Text deriving (Eq, Show)"
        ]
--- a/tests/Language/Haskell/Stylish/Step/Data/Tests.hs
+++ b/tests/Language/Haskell/Stylish/Step/Data/Tests.hs
@@ -30,6 +30,7 @@ tests = testGroup "Language.Haskell.Stylish.Step.Data.Tests"
     , testCase "case 17" case17
     , testCase "case 18" case18
     , testCase "case 19" case19
+    , testCase "case 20 (issue 262)" case20
     ]
 
 case00 :: Assertion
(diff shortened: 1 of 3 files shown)


tests/Language/Haskell/Stylish/Tests.hs
```diff
@@ -47,10 +46,9 @@ case02 = withTestDirTree $ do
     actual <- format (Just $ ConfigPath "test-config.yaml") Nothing input
     actual @?= result
   where
-    input = "module Herp where\n data Foo = Bar | Baz"
+    input = "module Herp where\ndata Foo = Bar | Baz"
     result = Right [ "module Herp where"
-                   , "data Foo = Bar"
-                   , "  | Baz"
+                   , "data Foo = Bar | Baz"
```
lukasz-golebiewski: `case01` and `case02` in this suite test whether the `indent` value was correctly read from the config file (and applied during formatting), or whether the default 4 was used. This behavior is no longer tested, so maybe it would make sense to add fields to `Bar` and `Baz` constructors?

tests/Language/Haskell/Stylish/Step/Data/Tests.hs
```diff
@@ -389,3 +385,15 @@ case19 = expected @=? testStep (step 2) input
...
+-- | Should not break Enums (data without records) formating
+--
+-- See https://github.com/jaspervdj/stylish-haskell/issues/262
+case20 :: Assertion
+case20 = input @=? testStep (step 2) input
+  where
+    input = unlines
+       [ "module Herp where"
+       , ""
+       , "data Tag = Title | Text"
```
lukasz-golebiewski:
```suggestion
       , "data Tag = Title | Text deriving (Eq)"
```

tests/Language/Haskell/Stylish/Tests.hs
maksbotan: Done!

danstn:
How soon will a version with this fix be released you reckon? 

Kinda broke styling for the whole codebase :)

jaspervdj:
@danbst Sorry, this fell off my radar!  I'll upload a new version today.

jaspervdj:
@danbst Released as `v0.11.0.0`!

danstn:
What a legend! Thank you @jaspervdj :)

PR #270, opened 2020-02-04
Optional switch, to ignore files/directories executing stylish-haskell

mmzx:
Adds command-line switch _-e_ to declare files to be excluded from executing stylish-haskell on them.
Adds command-line switch _-r_ that finds haskell source files in provided input directories.

Files/folders declared to be exceptions are additionally taken into account when the _-r_ recursive switch is enabled.

Request to skip files were made in #200 .
Directory recursing is mentioned in #162 .


src/Main.hs
```diff
@@ -21,14 +26,17 @@ import           Language.Haskell.Stylish
...
-    , saConfig   :: Maybe FilePath
-    , saVerbose  :: Bool
-    , saDefaults :: Bool
-    , saInPlace  :: Bool
-    , saNoUtf8   :: Bool
-    , saFiles    :: [FilePath]
-    } deriving (Show)
+    { saVersion   :: Bool
+    , saConfig    :: Maybe FilePath
+    , saRecursive :: Maybe FilePath
```
jaspervdj: I think this should be a boolean, and if it is set, we should treat the files in `saFiles` as the paths to recursively search.  Optionally, we can rename `saFiles` to `saPaths`?

mmzx: Thanks @jaspervdj ! :) 

- _-r_ takes a path, in case it  was a folder a recursive search would be performed and _hs_ files would be resulted as its output. So, this is not just something to be switched on or off.

- _-e_ takes a path, which may be a folder or a  file; to define blacklist of files not to be formatted by stylish-haskell

The two switches are not necessary used together.

Using _-r_ or _-e_ switches would contribute to change the content of ```saFiles``` which would store file pathes no matter what. Certainly we can rename ```saFiles``` to ```saPaths```.

mmzx:
> The code looks good to me but I'm not sure what the advantage of this is over something like:
> 
> ```
> find . -name '*.hs' -exec stylish-haskell -i '{}' \;
> ```
> 
> Specialized tools like `git ls-files`, `find`, ... will always allow better finding than we can possibly implement. I'm okay with adding some very simple recursion if you think it's useful, but I'm not sure if the filtering/exclude is worth it because at that point users may as well just switch to `find` & friends?

Depending on CI environment writing a detection of change may be easy or difficult. Recently I could have saved quiet a bit of time if I had this feature integrated into stylish-haskell... and the final solution wasn't very nice in that case.
So, having it at hand would make life a bit easier and could avoid writing e.g. checks in shell in some cases for sure.

mmzx:
Let me think it through again in the evening...

mmzx:
@jaspervdj , let me know if there are further things to be considered. If you would accept this, probably I would need to update the documentation as well.

mmzx:
> Thanks for this PR! I think there's currently two issues with the code.
> 
>     * The name `withExceptions` is quite bad, it suggests it has something to do with exceptions in Haskell (i.e. `Control.Exception`), but that's not the case.
Sure. I will come up with something around the idea of _excludes_. Naming things is difficult.

> 
>     * A bigger issue is that this doesn't seem like the right implementation.  One of the reasons you would want to exclude a directory is that it is waaay to big (i.e. `.git`) and traversing it slows down `stylish-haskell` to much to run it on every save.  The code you have now builds a full list of all files, which is problematic -- the excludes would need to be ignored within the directory tree traversing code.
I rather kept CI execution in mind and not the use-case of interactivity / fast feedback.
Thanks a lot, I will look into it...

lib/Language/Haskell/Stylish.hs
```diff
@@ -113,8 +120,17 @@ format maybeConfigPath maybeFilePath contents = do
 
 --------------------------------------------------------------------------------
 -- | Searches Haskell source files in any given folder recursively.
-findHaskellFiles :: Bool -> [FilePath] -> IO [FilePath]
-findHaskellFiles v fs = mapM (findFilesR v) fs >>= return . concat
+findHaskellFiles :: Bool -> [FilePath] -> ExcludeList -> IO [FilePath]
+findHaskellFiles v fs es =
+  mapM (findFilesR v) (excludeFiles fs es) >>= return . flip excludeFiles es . concat
```
mmzx: The reason to call `excludeFiles`
- _before_ performing directory traversing is to eliminate folders that are explicitly specified to be excluded, so recursing on them would be a waste of work anyway. E.g. large folders.
and
- _after_, so any explicitly specified file is to be excluded in a recursively traversed folder. E.g.: `stylish-haskell -r -e ./src/Bambi.hs ./src`

mmzx:
Bump.

PR #286, opened 2020-06-27
Abstraction over GHC parsing

jaspervdj:


jaspervdj:
Merged as part of #307

PR #293, opened 2020-07-23
Use ghc-lib-parser with stylish-haskell

felixmulder:
This PR contains the following:

- [x] Use ghc-lib-parser for Data
- [x] Add module header step
- [x] Add hardcoded imports step
- [x] Allows non-converted steps to be run alongside new ones
- [ ] Refactor imports step to use old `Options` configuration
- [x] Language pragmas step
- [ ] Align
- [ ] Squash step (completed by GSoC?)

This is at the point where I think we should start discussing the changes.

## Known issues
- The printer monad does not care about line length
- The imports are not deduplicated
- ~Does not support GADTs yet (should be quite trivial)~ **EDIT:** Basic GADT support is now included
- Comments within deriving clauses are hard to put back next to the correct clause. I've opted for placing them before all deriving clauses as an intermediate workaround. We should file an issue for this if we merge this.
- ~Haven't implemented getting rid of redundant language pragmas~

## Scope of PR
I'll be happy to fix most of these issues, but I'm not too keen on doing:

- Align step
- Squash step
- Refactoring imports to use `Options` datatype
- ~Implement removal of redundant language pragmas~

## Improvements to be had
Since this PR lays the foundation of having a context aware printer monad, we should be able to use it in order to have automatic line breaking etc at appropriate times. I think it's a huge improvement!


felixmulder:
Seems like stack moved its location of configuration files from `fpco` to `commercialhaskell`, I bumped the docker image to `8.8.3` since that contains a newer version of stack that was able to figure this out. Don't know how you feel about that, it might be possible to do `stack upgrade` as a step in the old image if you want the image to stay on `8.6.x`

jaspervdj:
This is awesome @felixmulder!  I'm currently on holiday, but I will start reviewing this tomorrow or Monday, and then I can talk to our GSoC student as well to plan how we'll integrate both branches.  Please don't worry about CI too much if it's giving you trouble, I've been planning to switch it over to GitHub actions.

felixmulder:
At this point I am starting to feel like it is good enough for review and potentially to merge as a good step in the right direction :v:

felixmulder:
Thanks @jaspervdj! I fixed all the CI issues now and so this should be in pretty good shape. Looking forward to your review!

src/Main.hs
```diff
@@ -5,9 +5,9 @@ module Main
 
 
 --------------------------------------------------------------------------------
-import           Control.Monad            (forM_, unless)
+import           Control.Monad            (forM_, unless, when)
 import qualified Data.ByteString.Char8    as BC8
-import           Data.Monoid              ((<>))
+--import           Data.Monoid              ((<>))
```
lukasz-golebiewski, line 10: nitpick: remove me

lib/Language/Haskell/Stylish/Config.hs
```diff
@@ -137,13 +152,18 @@ parseConfig (A.Object o) = do
...
+        [ ("normal", NormalExitBehavior)
+        , ("error_on_format", ErrorOnFormatExitBehavior)
+        ]
 parseConfig _            = mzero
 
 
 --------------------------------------------------------------------------------
 catalog :: Map String (Config -> A.Object -> A.Parser Step)
-catalog = M.fromList
+catalog = M.fromList $
```
lukasz-golebiewski, line 164: nitpick: is the $ necessary?

lib/Language/Haskell/Stylish/GHC.hs
```diff
@@ -0,0 +1,99 @@
...
+
+    llvmConfig = GHC.LlvmConfig [] []
+
+unLocated :: Located a -> a
+unLocated (L _ a) = a
+
+showOutputable :: GHC.Outputable a => a -> String
+showOutputable = GHC.showPpr baseDynFlags
+
+compareOutputable :: GHC.Outputable a => a -> a -> Ordering
```
lukasz-golebiewski, line 97: Do we have any other orderings in use? If not maybe it would make sense to extract this to an orphan `Ord` instance for `GHC.Outputable` and not use this one explicitly?

src/Main.hs
felixmulder: I was unsure of this one - I removed it since it was giving me warnings. But, I guess it's necessary for older version of base?

lib/Language/Haskell/Stylish/GHC.hs
felixmulder: This is mostly used in the imports step. I think it might be a good idea here to invent our type for imports that carries the information we need. This'll probably be nice when we add support for merging imports - so then we add the ordering instance for that type.

WDYT?

lukasz-golebiewski: Sounds good!

lib/Language/Haskell/Stylish/Step/Imports'.hs
```diff
@@ -0,0 +1,198 @@
+{-# LANGUAGE BlockArguments #-}
+{-# LANGUAGE DoAndIfThenElse #-}
+{-# LANGUAGE LambdaCase #-}
+module Language.Haskell.Stylish.Step.Imports'
```
jaspervdj, line 4: Having a file with `'` is unfortunately bound to cause trouble in loose `grep` / `find` / ... processing of the source code, we should rename this to `Imports2.hs` or `ImportsGHC.hs`.

src/Main.hs
felixmulder: I stubbed this out using the "standard CPP way"

lib/Language/Haskell/Stylish/Step/Imports'.hs
felixmulder: Fixed!

lib/Language/Haskell/Stylish/GHC.hs
felixmulder: So I did two things, I changed the newtype representation of:

```haskell
newtype Imports = Imports [LImportDecl GhcPs]

-- to:

newtype Import = Import (ImportDecl GhcPs)
```

This makes things a bit easier to reason about, it was possible to define equality for the import individually and I think it'll give us a more intuitive API down the line to work with imports. E.g. I added a helper function `canMergeImport` which is used in the equality instance. This helper could be used in the `Imports` step to merge imports - and get better ordering... I think I'm gonna play with that a bit today

jaspervdj:
@felixmulder I spoke to @VBeatrice and the current plan is for her to finish the Align step.  After that we can merge the Squash and Align steps from GSoC into your work.  Finally, we'll probably need to do some remaining bits (the `Options` you mentioned), and removing a bunch of dead code.  I can do together with the student and hopefully that'll put us in a good position to make a release.

felixmulder:
@jaspervdj - sounds good to me :+1:

There are a couple of systemic things that I think we should keep in mind moving forward:

Comments are a pain in the rear. I think we basically have two choices here:

- Make each step responsible for preserving comments (current approach)
- Put comment handling into the printer monad

The former puts the onus on the implementer to get comments right for their step. It _does_ give more precision and maybe better formatting(?).

It could be abstracted into the printer monad. Which makes it easier for the implementer, and arguably - if we want to move line-wrapping into the printer monad, it might make that job easier as well.

Not sure at this point, but I don't think we're painting ourselves into a corner either way. The testing covers things fairly well at this point and our monorepo has >50kLOC that we're running this branch of stylish on for each commit since a few weeks back.

lib/Language/Haskell/Stylish/GHC.hs
shayne-fletcher: @lukasz-golebiewski @felixmulder

> Do we have any other orderings in use? If not maybe it would make sense to extract this to an orphan `Ord` instance for `GHC.Outputable` and not use this one explicitly?

This thread caught my eye as I perused this ticket.  It looks to me that these lines are covering the same ground as [`ghc-lib-parser-ex`](https://hackage.haskell.org/package/ghc-lib-parser-ex)'s [`HsExtendInstances` wrapper](https://hackage.haskell.org/package/ghc-lib-parser-ex-8.10.0.16/docs/Language-Haskell-GhclibParserEx-GHC-Hs-ExtendInstances.html). I've [mused before](https://github.com/ndmitchell/hlint/issues/1004) that `ghc-lib-parser-ex` might be useful for stylish-haskell to use/extend on its conversion from haskell-src-exts to `ghc-lib-parser` - would be nice to share these sorts of common utilities with e.g. HLint!

shayne-fletcher: Hmm c.f. `showOutputable :: GHC.Outputable a => a -> String` with [`unsafePrettyPrint`](https://hackage.haskell.org/package/ghc-lib-parser-ex-8.10.0.16/docs/Language-Haskell-GhclibParserEx-GHC-Utils-Outputable.html#v:unsafePrettyPrint) too.

jaspervdj:
GSoC is over for this year.  I'll be working on merging in these changes and doing some final checks this week / weekend.

jaspervdj:
Merged in 250e7091edd93ce5a476706ddd968ef3ec1ef336, thanks again to everyone involved!

PR #299, opened 2020-08-12
Use `-split-sections`

fosskers:
This PR reduces the final binary size of `stylish-haskell` on my machine from 19mb to 7.9mb. Experimentally, using `-split-sections` like this doesn't introduce a compilation time penalty, although it does require that dependencies be rebuilt (hopefully once and then cached, as with CI).

I also bumped the Stackage LTS, and removed pins that were no longer necessary.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,5 +1,8 @@
 resolver: lts-16.9
 
+# Specifying `-split-sections` in this way propagates the setting to all
+# dependencies as well. The effect of this is a 50%-60% reduction in final
+# binary size, with effectively no additional compilation time cost.
 ghc-options:
   $everything: -split-sections
 
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,13 +1,12 @@
-resolver: lts-15.6
-packages:
-- '.'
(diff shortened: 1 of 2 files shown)


stack.yaml
```diff
@@ -1,13 +1,12 @@
-resolver: lts-15.6
-packages:
-- '.'
+resolver: lts-16.9
+
+ghc-options:
+  $everything: -split-sections
```
jaspervdj, line 4: Looks good to me, could you add a quick comment why the flag is used here as well?  Otherwise I'll probably forget.

fosskers: Sure, I will once I'm home. I should mention that it's not enough just to specify it in a project's `.cabal`, since that won't propagate the setting to its dependencies. Hence its done in the `stack.yaml`, targeting `$everything`.

fosskers: Done.

jaspervdj:
This was removed for now because of #304

fosskers:
Is there a way to specify `ghc-options` in a `stack.yaml` per platform?

Issue #304, opened 2020-08-22
stack build broken on macOS for use of split-sections


shayne-fletcher:
The use of
```
ghc-options:
  $everything: -split-sections
```
prevents `stack build` from succeeding on macOS.

PR #305, opened 2020-08-22
Remove split-sections stanza

shayne-fletcher:
The use of `split-sections` in stack.yaml breaks the build on macOS. Fixes https://github.com/jaspervdj/stylish-haskell/issues/304.


jaspervdj:
Thanks!

PR #307, opened 2020-09-14
Merge in ghc-lib-parser patches

jaspervdj:
This is an integration PR to merge #293 and the GSOC work.  There are still
a number of backwards-incompatible changes that need to be sorted out
before we can fully release this, and I'll keep notes of any significant changes
here.

# Changes

 -  [x] I changed the sorting order of imported things back to the following order:
    constructors/data types/classes, followed by functions, followed by operators.

 -  [x] I changed the grouping of imports to check for consecutive lines of imports
    rather than using comments as separators (but comments should still be
    preserved, of course).

 -  [x] The `Imports` step is not taking the `global` alignment option into account.

 -  [x] The printer code seems unequipped to do proper wrapping: it decides
    whether or not to print a newline by looking at the current line length, rather
    than the hypothetical line length if a comma, space and the next item are
    added.
 
    Looking at the printer code, it should be possible to add a combinator along
    the lines of:
 
        ifFitsOnLine
            :: P a  -- ^ Code that appends to the line, should not print a newline
            -> P a  -- ^ Code executed if the previous code surpasses line length
            -> P a  -- ^ Result of left or right branch

 -  [x] I think there is still an off-by-one error in the import wrapping code: it
    doesn't take the final `,` into account.

 -  [x] The style of consecutive lines for a single import has changed from
        
        import Data.List (init, last,
                          head)
 
    to:
 
        import Data.List (init, last)
        import Data.List (head)
 
    I'd like to revert the default back to the original but we'll want to retain the
    new style as an option.  Proposal is to add the new style as
    `list_align: repeat`.
 
    Note that this also involves adding all the other options here that `list_align`
    supports.

 -  [x] There are many more issues with various combinations of `long_list_align`,
    `list_align`, `space_surround` and `align`.

 -  [x] The `Align` step is having trouble with many files, for example:
 
        foo = align
          where
            align' = importAlign align
            padModuleNames' = padModuleNames align
            padNames = align' /= None && padModuleNames'
            padQual  = case align' of
              Global -> True
              File   -> fileAlign
              Group  -> anyQual
              None   -> False

 -  [x] The `ModuleHeader` step should allow configuring indentation

 -  [x] The `ModuleHeader` step should allow turning off sorting

 -  [x] The `ModuleHeader` step should use the same sorting order as `Imports`

 -  [x] The `LanguagePragmas` step is still using haskell-src-exts

 -  [x] The `UnicodeSyntax` step is still using haskell-src-exts

# Open issues


jaspervdj:
Merged in 250e7091edd93ce5a476706ddd968ef3ec1ef336, thanks again to everyone involved!

PR #311, opened 2020-10-04
Fix "group" import sort with multi-line imports

maksbotan:
When some import line spans multuple lines, e.g. when import list is
long, stylish-haskell breaks a group at this line, leading to bad
result.

This commits makes sure that import groups are recognized solely by
empty lines.

---

I hope it's straightforward from the test case :)
Thanks for this awesome tool. Most likely I will make a couple more PRs in the next few days.


jaspervdj (approved):

jaspervdj:
Thanks a lot!!

PR #317, opened 2020-10-05
Refactor UnicodeSyntax.hs

lukasz-golebiewski:


jaspervdj (approved):
Looks good!  One thing to note is that this really only works because all of the Unicode Syntax symbols we're converting are in type signatures?  It may be worth adding a note about that.

lukasz-golebiewski:
Are you suggesting adding a comment @jaspervdj ? Where exactly?

lib/Language/Haskell/Stylish/Step/UnicodeSyntax.hs
```diff
@@ -53,36 +53,15 @@ groupPerLine = M.toList . M.fromListWith (++) .
     map (\((r, c), x) -> (r, [(c, x)]))
```
jaspervdj, line 55:
```suggestion
-- | Find symbol positions in the module.  Currently only searches in type
-- signatures.
```

PR #325, opened 2020-10-09
Add `break_only_where` option

1Computer1:
Alternatively, make it the default behavior? I don't think I've ever seen someone add a new line before `where` like that before.

### DIFF
--- a/data/stylish-haskell.yaml
+++ b/data/stylish-haskell.yaml
@@ -30,6 +30,9 @@ steps:
   #
   #     # See `separate_lists` for the `imports` step.
   #     separate_lists: true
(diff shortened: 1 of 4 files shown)


lukasz-golebiewski:
Making it default makes more sense to me

1Computer1:
Sorry, do you mean to make
```hs
module Foo where
```
or
```hs
module Foo
  where
```
the default?

lukasz-golebiewski:
Hey! I agree with you, so let's make former the default :-)

lukasz-golebiewski (approved):
Nice!

PR #332, opened 2020-10-24
Better support for datakinds in records

jhmcstanton:
Addressing ticks dropping from records that contain promoted data types.


jhmcstanton:
Sorry for the awkward force push, missed that my commit message had not been cleaned up.

jaspervdj:
Thanks!

PR #337, opened 2020-10-30
Keep wildcard if present in IEThingWith

akrmn:
This is necessary to export pattern synonyms bundled with a type. I went with moving the ellipsis to the beginning of the `IEThingWith` list, to match the style from the GHC User's guide, though this could become a setting if there's interest for e.g. moving it to the end.

See https://downloads.haskell.org/ghc/latest/docs/html/users_guide/glasgow_exts.html#import-and-export-of-pattern-synonyms


jaspervdj:
Thanks a lot!

Issue #342, opened 2020-11-27
`newtype`s are formatted on two lines with `break_single_constructors: false`


imuli:
As `newtype`s always have a single constructor, you would think that they
would obey `break_single_constructors`. However, with that set to false,
```haskell
newtype Foo = Foo Int
```
becomes
```haskell
newtype Foo
   = Foo Int
```

while `data` with a single constructor stay on one line.
```haskell
data Foo = Foo Int
```

imuli:
Okay, having now found the place in the source that explicitly checks this _and_ the test case that claims the current behavior is correct, it seems like @jaspervdj wants this behavior. I still find it unintuitive - if it's not too much configuration bloat could we add a `break_newtypes` flag?

EncodePanda:
I think this accident @imuli . newtype and data has to be consistent. let's fix that.

stevladimir:
Hello, any progress on that? Maybe I can help somehow?

jaspervdj:
Yeah this is accidental.  I would welcome a PR that fixes this :-)

stevladimir:
IMO this should be the default behavior, but It is somewhat breaking change (people might get big diffs) if we just directly alter the current behavior. If this is acceptable I would go this way as I see no strong reasons to distinguish `newtype` and `data` in this respect.

Otherwise, @imuli has a [fix](https://github.com/haskell/stylish-haskell/issues/342) which should be backwards compatible in cost of introducing a new option.

Which one would you prefer?

imuli:
I think the primary issue with my fix as it stands is that it doesn't provide any tests?

I also would have no complaints about somone removing the option in favor of `break_single_constructors`.

PR #343, opened 2020-11-27
Add break_newtypes flag. (#342)

imuli:
Default behavior remains the same, but when you set `break_newtypes` to
`false` newtypes will be formatted on a single line.


imuli:
Happy to add tests if you all want this.

EncodePanda:
Yeah, this is great. Can you please provide a test for this?

PR #348, opened 2021-01-02
Add HLS as a way for editor integration (#347)

### LINKED ISSUE #347: Mentioning HLS plugin in README.markdown
For [Haskell Language Server(HLS)](https://github.com/haskell/haskell-language-server) supports stylish-haskell and I believe it is one of the most accessible ways to use style formatters in editors, it would be great if HLS is mentioned somewhere in README.markdown. Maybe as a part of editor support section?

Ailrun:
Fixes #347

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -158,7 +158,21 @@ data Foo a
   deriving (ToJSON) via Bar Foo
 ```
 
-## VIM integration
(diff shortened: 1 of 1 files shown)


jaspervdj:
Thanks!

PR #349, opened 2021-01-26
Generalise break_only_where to allow single-line module headers

mrBliss:
Previously, the `break_only_where` option of `module_header` allowed choosing
between the following two styles:

1. `break_only_where: false`:

   ```haskell
   module Foo where

   module Bar
       ( main
       ) where
   ```

2. `break_only_where: true`:

   ```haskell
   module Foo
       where

   module Bar
       ( main
       ) where
   ```

In case a module only has a single export (e.g., `main` or `tests`) or no
exports at all (e.g., a module defining orphans), we still "break" the `where`:

```haskell
module Bar
    ( main
    ) where
module Orphans
    (
    ) where
```

It should be possible to print these module headers as:

```haskell
module Bar (main) where
module Orphans () where
```

We generalise the `break_only_where` option to `break_where` with the following
three possible values:

- `exports`: only break when there is an explicit export list.
- `inline`: only break when the export list is too long. This is determined by
   the `columns` setting. Not applicable when the export list contains comments
   as newlines will be required.
- `always`: always break before the `where`.

### DIFF
--- a/data/stylish-haskell.yaml
+++ b/data/stylish-haskell.yaml
@@ -31,8 +31,15 @@ steps:
   #     # See `separate_lists` for the `imports` step.
   #     separate_lists: true
   #
-  #     # Whether to break the "where" if there are no exports.
-  #     break_only_where: false
+  #     # When to break the "where".
+  #     # Possible values:
+  #     # - exports: only break when there is an explicit export list.
+  #     # - single: only break when the export list counts more than one export.
+  #     # - inline: only break when the export list is too long. This is
+  #     #   determined by the `columns` setting. Not applicable when the export
+  #     #   list contains comments as newlines will be required.
+  #     # - always: always break before the "where".
+  #     break_where: exports
 
   # Format record definitions. This is disabled by default.
   #
--- a/lib/Language/Haskell/Stylish/Config.hs
+++ b/lib/Language/Haskell/Stylish/Config.hs
@@ -196,13 +196,22 @@ parseEnum strs _   (Just k) = case lookup k strs of
 
 --------------------------------------------------------------------------------
 parseModuleHeader :: Config -> A.Object -> A.Parser Step
-parseModuleHeader _ o = fmap ModuleHeader.step $ ModuleHeader.Config
-    <$> o A..:? "indent"           A..!= ModuleHeader.indent         def
-    <*> o A..:? "sort"             A..!= ModuleHeader.sort           def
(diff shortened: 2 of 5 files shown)


data/stylish-haskell.yaml
```diff
@@ -31,8 +31,14 @@ steps:
...
-  #     # Whether to break the "where" if there are no exports.
-  #     break_only_where: false
+  #     # When to break the "where".
+  #     # Possible values:
+  #     # - exports: only break when there is an explicit export list.
+  #     # - inline: only break when the export list is too long. This is
+  #     #   determined by the `columns` setting. Not applicable when the export
+  #     #   list contains comments as newlines will be required.
+  #     # - always: always break before the "where".
+  #     break_where: exports
```
mrBliss, line 41: If you can think of better names for these (or for `break_where`), please suggest them :slightly_smiling_face:

data/stylish-haskell.yaml
```diff
@@ -31,8 +31,14 @@ steps:
...
   #     separate_lists: true
   #
-  #     # Whether to break the "where" if there are no exports.
-  #     break_only_where: false
+  #     # When to break the "where".
+  #     # Possible values:
+  #     # - exports: only break when there is an explicit export list.
+  #     # - inline: only break when the export list is too long. This is
+  #     #   determined by the `columns` setting. Not applicable when the export
+  #     #   list contains comments as newlines will be required.
```
mrBliss, line 39: If desired, I can add a `space_surround` option to toggle between the following two styles:
```haskell
module Foo () where
module Foo (main) where
module Foo (main, bar) where
```
```haskell
module Foo ( ) where -- or () ?
module Foo ( main ) where
module Foo ( main, bar ) where
```
The option would only apply in case of `inline`.

mrBliss: Instead of `inline` (or in addition to?), I could also have something like `single`, which uses single-line formatting only when there are 0 to 1 exports. Starting from 2, multi-line formatting will be used.

data/stylish-haskell.yaml
mrBliss: Also: is backwards-compatibility important?

lukasz-golebiewski (approved):
This looks great, thanks!

lukasz-golebiewski:
@jaspervdj do you have any idea why the mac OS build is failing?

PR #357, opened 2021-03-09
Add new configuration to ModuleHeader step

EncodePanda:
Previously the formatter of module header would always add open
bracket of the export list on the next line after the module name.

With this change we give user ability to customize this behavior. The
open bracket ends up either on the same line as module name or on the
next line.

For backward compatibility we keep the 'next_line' a default behavior.

### DIFF
--- a/data/stylish-haskell.yaml
+++ b/data/stylish-haskell.yaml
@@ -40,6 +40,13 @@ steps:
   #     #   list contains comments as newlines will be required.
   #     # - always: always break before the "where".
   #     break_where: exports
+  #
+  #     # Where to put open bracket
+  #     # Possible values:
+  #     # - same_line: put open bracket on the same line as the module name, before the
+  #     #              the comment of the module
+  #     # - next_line: put open bracket on the next line, after module comment
+  #     open_bracket: next_line
 
   # Format record definitions. This is disabled by default.
   #
--- a/data/stylish-haskell.yaml
+++ b/data/stylish-haskell.yaml
@@ -44,7 +44,7 @@ steps:
   #     # Where to put open bracket
   #     # Possible values:
   #     # - same_line: put open bracket on the same line as the module name, before the
-  #     #              the comment of the module
+  #     #              comment of the module
   #     # - next_line: put open bracket on the next line, after module comment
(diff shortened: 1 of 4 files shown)


data/stylish-haskell.yaml
```diff
@@ -40,6 +40,13 @@ steps:
   #     #   list contains comments as newlines will be required.
   #     # - always: always break before the "where".
   #     break_where: exports
+  #
+  #     # Where to put open bracket
+  #     # Possible values:
+  #     # - same_line: put open bracket on the same line as the module name, before the
+  #     #              the comment of the module
```
lukasz-golebiewski, line 47: One `the` is redundant

lib/Language/Haskell/Stylish/Step/ModuleHeader.hs
```diff
@@ -221,6 +257,11 @@ printMultiLineExportList conf (L srcLoc exportsWithComments) = do
     -- > xxxx( -- Some comment
     -- > xxxxyyfoo
     -- > xxxx) where
+
+    firstChar =
+      if openBracket conf == SameLine then " "
```
lukasz-golebiewski, line 262: I would suggest using:
```haskell
case openBracket conf of
  SameLine -> " "
  NextLine -> "("
```
which is going to stop compiling when a new `OpenBracket` option is added

lukasz-golebiewski (approved):
Looks great! Just a few minor remarks

The number of test cases is starting to explode. We should think how to improve this.

data/stylish-haskell.yaml
EncodePanda: 👍

lib/Language/Haskell/Stylish/Step/ModuleHeader.hs
EncodePanda: good one, thx!

EncodePanda:
@lukasz-golebiewski regarding the number of tests comment, a number of tests will be always the number-of-initial-states (inputs) times the cardinality-of-the-configuration. It will grow exponentially I fear.

PR #365, opened 2021-04-10
Bump CI Mac Cache

EncodePanda:
Should we continue maintaining the Mac OS in CI? It's the second time I'm bumping it this week

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
       name: Cache ~/.stack
       with:
         path: ~/.stack
-        key: ${{ runner.os }}-${{ matrix.ghc }}-v5
+        key: ${{ runner.os }}-${{ matrix.ghc }}-v6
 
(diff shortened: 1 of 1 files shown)


jaspervdj:
@EncodePanda I don't have strong feelings about having Mac OS in CI, but I know some people are using the release binaries for Mac OS.  And, well, if we don't build Mac OS on each run, it can be a bit stressful if it turns out it's broken right before a release.

We could look into other options for the CI though, I think <https://github.com/haskell/actions> has been gaining some traction, or we could use [hashFiles](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#example-with-multiple-patterns) on the cabal/stack files to avoid the manual version bumps?

EncodePanda:
@jaspervdj There maybe could also be value in just relying on nix to build stylish. I think there are github actions allowing you to install Nix. We could then rely on caching mechanism (from IOHK's hydra to get the right GHC and from cachix as well - they have some plans for open source projects).

PR #379, opened 2021-08-28
Deriving clauses are not always on separate lines

stevemao:
So in the documentation:

> How many spaces to insert before "deriving" clause. Deriving clauses are always on separate lines.

But this case it is not. How to force deriving to be always on a separate line?


EncodePanda:
As @Lev135 mentioned, with his recent PR #424 this behavior is now fixed. We will cut new version soon I think, right @jaspervdj?

EncodePanda:
Fixed in #424

PR #383, opened 2021-11-05
test case-insensitive imports sort

vlatkoB:
One other test also fails:
```
Language.Haskell.Stylish.Config:
  Extensions extracted correctly from .cabal file: [Failed]
```


jhrcek:
Seems like this has been resolved, so the issue can be closed. Right?

Issue #397, opened 2022-04-18
nix-shell fails because cabal could not resolve dependencies


TikhonJelvis:
I cloned the repo and ran `nix-shell` without changing anything. It failed with an error. Here are the bits of the build log that seem relevant:

```
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: stylish-haskell-0.14.1.0 (user goal)
[__1] next goal: ghc-lib-parser-ex (dependency of stylish-haskell)
[__1] rejecting: ghc-lib-parser-ex-9.0.0.4 (conflict: stylish-haskell =>
ghc-lib-parser-ex>=9.2.0.3 && <9.3)
[__1] skipping: ghc-lib-parser-ex-9.0.0.3, ghc-lib-parser-ex-9.0.0.2,
```
...

```
ghc-lib-parser-ex-0.20200102, ghc-lib-parser-ex-8.10.0.18 (has the same
characteristics that caused the previous version to fail: excluded by
constraint '>=9.2.0.3 && <9.3' from 'stylish-haskell')
[__1] fail (backjumping, conflict set: ghc-lib-parser-ex, stylish-haskell)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: stylish-haskell, ghc-lib-parser-ex
```
...
```
error: attribute 'shellFor' missing

       at /nix/store/i05lrdjk7ifck3zkwdq2v9swxyf5m532-source/overlays/haskell.nix:545:35:

          544|             let project = project' // { recurseForDerivations = false; };
          545|             in rawProject // rec {
             |                                   ^
          546|               hsPkgs = final.lib.mapAttrs (packageName: package':
(use '--show-trace' to show detailed location information)
```

I'll give a try to debugging this. I haven't seriously used `haskell.nix` before, so I'm assuming the Nix error is caused by the `cabal` issue, but I'm not entirely sure.

andreasabel:
@TikhonJelvis : Is this issue fixed now or still open?

TikhonJelvis:
Pretty sure it was fixed in #398. That was a while ago so I don't remember the details and I'm not in a position to test it at the moment, but if it comes up again I can either reopen this issue or open a new one as appropriate.

andreasabel:
Good, then please close this issue.

PR #398, opened 2022-04-18
Fix Nix build, bumping haskell.nix (#397)

TikhonJelvis:
Fixes `nix-shell` and `nix-build` for the repo, which was failing with a dependency error (#397).

This PR takes two steps to fix this:

  1. It bumps the version of `haskell.nix` we use, switching to GHC 8.10.7 and Nixpkgs 21.11 so that we still use the haskell.nix cache

  2. After 1., I ran into an error where haskell.nix was using Cabal-3.2.1.0 even though `stylish-haskell.cabal` needs Cabal >= 3.4. To fix this, I had to explicitly mark the Cabal library as "reinstallable"—otherwise, haskell.nix uses the exact version that comes with GHC. (See [haskell-nix issue #1337][1] for details.)

[1]: https://github.com/input-output-hk/haskell.nix/issues/1337

I tested this change locally. Both `nix-build` and `nix-shell` succeeded without needing to build GHC from source.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,10 +1,26 @@
 { pkgs ? import ./haskell-pkgs.nix
-, haskellCompiler ? "ghc8104"
+, haskellCompiler ? "ghc8107"
 }:
 pkgs.haskell-nix.cabalProject {
   src = pkgs.haskell-nix.haskellLib.cleanGit {
     name = "stylish-haskell";
     src = ./.;
   };
+
   compiler-nix-name = haskellCompiler;
+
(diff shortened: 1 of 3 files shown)


TikhonJelvis:
@jaspervdj would you mind merging this? I can work around the Nix issues locally, but it's more convenient if it works from the `main` branch.

Unless I'm missing something, there is no way for this PR to affect anybody except for people using Nix for local development—should be low risk to merge :)

Thanks!

PR #399, opened 2022-04-19
Option to group imports automatically

TikhonJelvis:
This PR adds an option called `group_imports` (default: `false`) that automatically organizes imports into groups based on the first part of the module name:

``` haskell
import Data.List
import Control.Monad
import GHC.Exts
import Control.Applicative
import Data.Map
```

⇒

``` haskell
import Control.Applicative
import Control.Monad

import Data.List
import Data.Map

import GHC.Exts
```

I currently do this manually because I find that breaking up imports makes them much easier to scan. Now that I've started using haskell-language-server to insert most of my imports, it would be great if stylish-haskell could handle the grouping for me.

This is my first stab at the code to do this—not sure if there's a better approach.

Another option would be to let the user configure which order groups go in. Something like:

``` yaml
group_imports: ["*", "Test"]
```

which would have all groups except for `Test.*` first and `Test.*` last. The logic here would be similar to [scalafmt][1] but without regexps. This would let me configure the logic I use manually today, where I group library modules first, then test modules and finally modules internal to the current project. However, I am not sure whether this design would make sense to anybody except me and whether it's worth the added code complexity.

[1]: https://scalameta.org/scalafmt/docs/configuration.html#imports-groups

### DIFF
--- a/data/stylish-haskell.yaml
+++ b/data/stylish-haskell.yaml
@@ -294,6 +294,107 @@ steps:
       # Default: false
       post_qualify: false
 
+      # Automatically group imports based on their module names, with
+      # a blank line separating each group. Groups are ordered in
+      # alphabetical order.
+      #
+      # By default, this groups by the first part of each module's
+      # name (Control.* will be grouped together, Data.*... etc), but
+      # this can be configured with the group_patterns setting.
+      #
+      # When enabled, this rewrites existing blank lines and groups.
+      #
+      # - true: Group imports by the first part of the module name.
+      #
+      #   > import Control.Applicative
+      #   > import Control.Monad
+      #   > import Control.Monad.MonadError
+      #   >
+      #   > import Data.Functor
+      #
+      # - false: Keep import groups as-is (still sorting and
+      #   formatting the imports within each group)
+      #
+      #   > import Control.Monad
+      #   > import Data.Functor
+      #   >
+      #   > import Control.Applicative
+      #   > import Control.Monad.MonadError
+      #
+      # Default: false
+      group_imports: false
+
+      # A list of rules specifying how to group modules and how to
+      # order the groups.
+      #
+      # Each rule has a match field; the rule only applies to module
+      # names matched by this pattern. Patterns are POSIX extended
+      # regular expressions; see the documentation of Text.Regex.TDFA
+      # for details:
+      # https://hackage.haskell.org/package/regex-tdfa-1.3.1.2/docs/Text-Regex-TDFA.html
+      #
+      # Rules are processed in order, so only the *first* rule that
+      # matches a specific module will apply. Any module names that do
+      # not match a single rule will be put into a single group at the
+      # end of the import block.
+      #
+      # Example: group MyApp modules first, with everything else in
+      # one group at the end.
+      #
+      #  group_rules:
+      #    - match: "^MyApp\\>"
+      #
+      #  > import MyApp
+      #  > import MyApp.Foo
+      #  >
+      #  > import Control.Monad
+      #  > import MyApps
+      #  > import Test.MyApp
+      #
+      # A rule can also optionally have a sub_group pattern. Imports
+      # that match the rule will be broken up into further groups by
+      # the part of the module name matched by the sub_group pattern.
+      #
+      # Example: group MyApp modules first, then everything else
+      # sub-grouped by the first part of the module name.
+      #
+      #  group_rules:
+      #    - match: "^MyApp\\>"
+      #    - match: "."
+      #      sub_group: "^[^.]+"
+      #
+      #  > import MyApp
+      #  > import MyApp.Foo
+      #  >
+      #  > import Control.Applicative
+      #  > import Control.Monad
+      #  >
+      #  > import Data.Map
+      #
+      # A pattern only needs to match part of the module name, which
+      # could be in the middle. You can use ^pattern to anchor to the
+      # beginning of the module name, pattern$ to anchor to the end
+      # and ^pattern$ to force a full match. Example:
+      #
+      #  - "Test\\." would match "Test.Foo" and "Foo.Test.Lib"
+      #  - "^Test\\." would match "Test.Foo" but not "Foo.Test.Lib"
+      #  - "\\.Test$" would match "Foo.Test" but not "Foo.Test.Lib"
+      #  - "^Test$" would *only* match "Test"
+      #
+      # You can use \\< and \\> to anchor against the beginning and
+      # end of words, respectively. For example:
+      #
+      #  - "^Test\\." would match "Test.Foo" but not "Test" or "Tests"
+      #  - "^Test\\>" would match "Test.Foo" and "Test", but not
+      #    "Tests"
+      #
+      # The default is a single rule that matches everything and
+      # sub-groups based on the first component of the module name.
+      #
+      # Default: [{ "match" : ".*", "sub_group": "^[^.]+" }]
+      group_rules:
+        - match: ".*"
+          sub_group: "^[^.]+"
 
   # Language pragmas
   - language_pragmas:
--- a/lib/Language/Haskell/Stylish/Step/Imports.hs
+++ b/lib/Language/Haskell/Stylish/Step/Imports.hs
@@ -1,7 +1,10 @@
-{-# LANGUAGE BlockArguments  #-}
-{-# LANGUAGE DoAndIfThenElse #-}
-{-# LANGUAGE LambdaCase      #-}
-{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE BlockArguments    #-}
+{-# LANGUAGE DoAndIfThenElse   #-}
+{-# LANGUAGE FlexibleContexts  #-}
+{-# LANGUAGE LambdaCase        #-}
+{-# LANGUAGE NamedFieldPuns    #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards   #-}
 module Language.Haskell.Stylish.Step.Imports
   ( Options (..)
   , defaultOptions
@@ -10,22 +13,31 @@ module Language.Haskell.Stylish.Step.Imports
   , LongListAlign (..)
   , EmptyListAlign (..)
   , ListPadding (..)
+  , GroupRule (..)
   , step
 
   , printImport
+
+  , parsePattern
+  , unsafeParsePattern
   ) where
 
 --------------------------------------------------------------------------------
 import           Control.Monad                     (forM_, void, when)
+import qualified Data.Aeson                        as A
 import           Data.Foldable                     (toList)
 import           Data.Function                     (on, (&))
 import           Data.Functor                      (($>))
-import           Data.List                         (sortBy)
+import           Data.List                         (groupBy, intercalate,
+                                                    partition, sortBy, sortOn)
 import           Data.List.NonEmpty                (NonEmpty (..))
 import qualified Data.List.NonEmpty                as NonEmpty
 import qualified Data.Map                          as Map
-import           Data.Maybe                        (fromMaybe, isJust)
+import           Data.Maybe                        (fromMaybe, isJust, mapMaybe)
+import           Data.Sequence                     (Seq ((:|>)))
+import qualified Data.Sequence                     as Seq
 import qualified Data.Set                          as Set
+import qualified Data.Text                         as T
 import qualified GHC.Data.FastString               as GHC
 import qualified GHC.Hs                            as GHC
 import qualified GHC.Types.Name.Reader             as GHC
@@ -33,7 +45,9 @@ import qualified GHC.Types.SourceText              as GHC
 import qualified GHC.Types.SrcLoc                  as GHC
 import qualified GHC.Unit.Module.Name              as GHC
 import qualified GHC.Unit.Types                    as GHC
-
+import           Text.Regex.TDFA                   (Regex)
+import qualified Text.Regex.TDFA                   as Regex
+import           Text.Regex.TDFA.ReadRegex         (parseRegex)
 
 --------------------------------------------------------------------------------
 import           Language.Haskell.Stylish.Block
@@ -43,7 +57,6 @@ import           Language.Haskell.Stylish.Ordering
 import           Language.Haskell.Stylish.Printer
 import           Language.Haskell.Stylish.Step
 import           Language.Haskell.Stylish.Util
-
 
 --------------------------------------------------------------------------------
 data Options = Options
@@ -56,20 +69,28 @@ data Options = Options
     , separateLists  :: Bool
     , spaceSurround  :: Bool
     , postQualified  :: Bool
+    , groupImports   :: Bool
+    , groupRules     :: [GroupRule]
     } deriving (Eq, Show)
 
 defaultOptions :: Options
 defaultOptions = Options
-    { importAlign     = Global
-    , listAlign       = AfterAlias
-    , padModuleNames  = True
-    , longListAlign   = Inline
-    , emptyListAlign  = Inherit
-    , listPadding     = LPConstant 4
-    , separateLists   = True
-    , spaceSurround   = False
-    , postQualified   = False
+    { importAlign    = Global
+    , listAlign      = AfterAlias
+    , padModuleNames = True
+    , longListAlign  = Inline
+    , emptyListAlign = Inherit
+    , listPadding    = LPConstant 4
+    , separateLists  = True
+    , spaceSurround  = False
+    , postQualified  = False
+    , groupImports   = False
+    , groupRules     = [defaultGroupRule]
     }
+  where defaultGroupRule = GroupRule
+          { match    = unsafeParsePattern ".*"
+          , subGroup = Just $ unsafeParsePattern "^[^.]+"
+          }
 
 data ListPadding
     = LPConstant Int
@@ -103,6 +124,77 @@ data LongListAlign
     | Multiline -- multiline
     deriving (Eq, Show)
 
+-- | A rule for grouping imports that specifies which module names
+-- belong in a group and (optionally) how to break them up into
+-- sub-groups.
+--
+-- See the documentation for the group_rules setting in
+-- data/stylish-haskell.yaml for more details.
+data GroupRule = GroupRule
+  { match    :: Pattern
+    -- ^ The pattern that determines whether a rule applies to a
+    -- module name.
+  , subGroup :: Maybe Pattern
+    -- ^ An optional pattern for breaking the group up into smaller
+    -- sub-groups.
+  } deriving (Show, Eq)
+
+instance A.FromJSON GroupRule where
+  parseJSON = A.withObject "group_rule" parse
+    where parse o = GroupRule
+                <$> (o A..: "match")
+                <*> (o A..:? "sub_group")
+
+-- | A compiled regular expression. Provides instances that 'Regex'
+-- does not have (eg 'Show', 'Eq' and 'FromJSON').
+--
+-- Construct with 'parsePattern' to maintain the invariant that
+-- 'string' is the exact regex string used to compile 'regex'.
+data Pattern = Pattern
+  { regex  :: Regex
+    -- ^ The compiled regular expression.
+  , string :: String
+    -- ^ The valid regex string that 'regex' was compiled from.
+  }
+
+instance Show Pattern where show = show . string
+
+instance Eq Pattern where (==) = (==) `on` string
+
+instance A.FromJSON Pattern where
+  parseJSON = A.withText "regex" parse
+    where parse text = case parsePattern $ T.unpack text of
+            Left err  -> fail $ "Invalid regex:\n" <> err
+            Right pat -> pure pat
+
+
+-- | Parse a string into a compiled regular expression ('Pattern').
+--
+-- Returns a human-readable parse error message if the string is not
+-- valid regex syntax.
+--
+-- >>> parsePattern "^([^.]+)"
+-- Right "^([^.]+)"
+--
+-- >>> parsePattern "("
+-- Left "\"(\" (line 1, column 2):\nunexpected end of input\nexpecting empty () or anchor ^ or $ or an atom"
+parsePattern :: String -> Either String Pattern
+parsePattern string = case parseRegex string of
+  Right _  -> Right $ Pattern { string, regex = Regex.makeRegex string }
+  Left err -> Left (show err)
+
+-- | Parse a string into a regular expression, raising a runtime
+-- exception if the string is not valid regex syntax.
+--
+-- >>> unsafeParsePattern "^([^.]+)"
+-- "^([^.]+)"
+--
+-- >>> unsafeParsePattern "("
+-- "*** Exception: "(" (line 1, column 2):
+-- unexpected end of input
+-- expecting empty () or anchor ^ or $ or an atom
+unsafeParsePattern :: String -> Pattern
+unsafeParsePattern = either error id . parsePattern
 
 --------------------------------------------------------------------------------
 step :: Maybe Int -> Options -> Step
@@ -111,11 +203,15 @@ step columns = makeStep "Imports (ghc-lib-parser)" . printImports columns
 
 --------------------------------------------------------------------------------
 printImports :: Maybe Int -> Options -> Lines -> Module -> Lines
-printImports maxCols align ls m = Editor.apply changes ls
+printImports maxCols options ls m = Editor.apply changes ls
   where
     groups = moduleImportGroups m
     moduleStats = foldMap importStats . fmap GHC.unLoc $ concatMap toList groups
-    changes = foldMap (formatGroup maxCols align moduleStats) groups
+    changes
+      | groupImports options =
+          groupAndFormat maxCols options moduleStats groups
+      | otherwise =
+          foldMap (formatGroup maxCols options moduleStats) groups
 
 formatGroup
     :: Maybe Int -> Options -> ImportStats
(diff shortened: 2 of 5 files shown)


jaspervdj:
Thanks for the PR @TikhonJelvis!

I think something like this makes sense.  One thing I'm wondering about is that a for some libraries, taking 1 level makes little sense (i.e. some libraries are entirely nested under `Data.`, `Text.` or `Control.`).  Should this be configurable?  I feel like it heavily depends on the libraries you are importing, and you may not want the same level in each instance.

TikhonJelvis:
Yeah, that makes sense. Seems like it could fit into the “alternate configuration” scheme I described, but I'd have to think through the details a bit.

Here's a first stab at it: let's say that we want to group `Control.Lens` together separately from `Control.*` and also order things so that the `Test.*` is always last no matter what. Here's one idea of how to specify this:

``` yaml
group_imports: ["*", "Control.Lens.*", "*", "Test.*"]
```

This setting makes sense to me *intuitively*, but the details end up a bit weird. In particular:

  * `Foo.*` would cover `Foo` as well (which seems fine?)
  * multiple instances of the same level (eg multiple `*`) would get split up based on the alphabetic ordering of the first group between them
  
This second rule feels pretty *ad hoc* though. An alternate option would be to separate grouping and sorting:

``` yaml
group_imports:
  extra_groups: ["Control.Lens.*"]
  sort: ["*", "Test.*"]
```

Do either of these approaches make sense to you?

TikhonJelvis:
Was just thinking about this—a third approach would be to follow the same pattern as `scalafmt` *including* using regular expressions instead of our own matching syntax. The upside is that this would be simpler to implement than my other suggestions *and* more flexible. The downside is that it might be a bit less convenient for people who don't already know how to write regular expressions.

Originally I was also worried that we'd have to add a regex library and its transitive dependencies to stylish-haskell, but I just checked and stylish-haskell already (indirectly) depends on `regex-base` and `regex-posix`, so depending on those libraries directly would not be a big change.

After thinking through the previous two options I had, I'd probably lean towards this third option instead—this way the logic inside stylish-haskell itself could be reasonably simple and, if people want to do something off the beaten path, they can craft a moderately complex regular expression to achieve that. However, I don't have a great intuition for what settings would make sense here or how this fits into the overall design of stylish-haskell, so would love some design suggestions/feedback.

[1]: https://scalameta.org/scalafmt/docs/configuration.html#imports-groups

michaelpj:
Just to chime in as another potential user. My typical manual pattern is
- First all the imports from the current package
- Then everything else

Which suggests that I'd probably want both control over sorting and control over blocks... I think I'd use this feature regardless, but if we _can_ get that flexibility it would be nice!

TikhonJelvis:
Glad to hear that :).

Now that I've thought about the design a bit, I'm leaning towards a regular-expression-based system. The design still feels a bit fiddly and complex, but it covers most of what I'd want and I can't think of anything better—would love any suggestions.

I'll start implementing the logic soon; for now, here's the documentation comment I've written out for the `group_patterns` YAML configuration option:

> A list of regular expressions specifying how to group modules
> and how to order the groups.
> 
> Each pattern defines a single group. An import belongs to the
> first group where the pattern matches the name of the module
> being imported. Matches can be partial; to require matching
> the entire module name, you can use `^` and `$` in the pattern (eg
> `"^pattern$"`).
> 
> Example: `["Test\\.", "Foo\\."]` will put all test imports into
> a single group, followed by a group for `Foo.*`, followed by
> anything else:
> 
> ``` haskell
> import Test.Foo.A
> import Test.Foo.B
> import Test.Bar
> import Test.Baz.Qux
> 
> import Foo.A
> import Foo.B
> 
> import Control.Applicative
> import Data.Map
> ```
> 
> If a pattern uses a capture group, imports that match the
> pattern are broken up into further groups based on the
> captured text. The sub-groups are sorted relative to each
> other in alphabetical order by the captured text.
> 
> Example: `"Test\\.([^.]+)"` will produce sub-groups as follows:
> 
> ``` haskell
> import Test.Bar
> 
> import Test.Baz.Qux
> 
> import Test.Foo.A
> import Test.Foo.B
> ```
> 
> Any imports that do not match a pattern will be combined at a
> group at the end of the import block.
> 
> The default is `["([^.])+"]` which means "group everything based
> on the first component of the module name" (`Control.*`
> together, `Data.*` together... etc).

TikhonJelvis:
Okay, I implemented the regex-based grouping logic. It isn't 100% flexible, but it can cover patterns like "put internal imports first" and "group `Data.*` by the second level of the namespace". On the other hand, things like "put internal imports last" as well as generally having fine-grained control over ordering is either awkward or impossible, but I haven't been able to come up with a  better design.

I started out trying to use [`regex-posix`][3] because it was already in `stylish-haskell`'s transitive dependencies, but it turns out that `regex-posix` doesn't support Unicode (as well as being reportedly [slow and buggy][2]). I ended up using [`regex-tdfa`][1] instead, which seems like a nice library written in pure Haskell and without too many additional dependencies of its own.

Would love any suggestions on improving either the behavior or the (somewhat fiddly) code :)

[1]: https://hackage.haskell.org/package/regex-tdfa

[2]: https://wiki.haskell.org/Regex_Posix

[3]: https://hackage.haskell.org/package/regex-posix-0.96.0.1

data/stylish-haskell.yaml
```diff
@@ -294,6 +294,83 @@ steps:
...
+      # A list of regular expressions specifying how to group modules
+      # and how to order the groups.
+      #
+      # Each pattern defines a single group. An import belongs to the
+      # first group where the pattern matches the name of the module
+      # being imported. Matches can be partial; to require matching
+      # the entire module name, you can use ^ and $ in the pattern (eg
+      # "^pattern$").
+      #
+      # Example: ["Test\\.", "Foo\\."] will put all test imports into
```
michaelpj, line 336: Maybe include some tricky examples for this case, e.g. I believe `Foo.Test.Lib` will end up in the first block but `Foo.Test` will end up in the second! Or perhaps you should put `^`s in the regexes :)

data/stylish-haskell.yaml
```diff
@@ -294,6 +294,83 @@ steps:
...
+      #  > import Test.Bar
+      #  > import Test.Baz.Qux
+      #  >
+      #  > import Foo.A
+      #  > import Foo.B
+      #  >
+      #  > import Control.Applicative
+      #  > import Data.Map
+      #
+      # If a pattern uses a capture group, imports that match the
```
michaelpj, line 351: what if there is more than one capture group?

data/stylish-haskell.yaml
```diff
@@ -294,6 +294,83 @@ steps:
...
+      #   > import Test.Foo.B
+      #
+      # Any imports that do not match a pattern will be combined at a
+      # group at the end of the import block.
+      #
+      # The default is ["([^.])+"] which means "group everything based
+      # on the first component of the module name" (Control.*
+      # together, Data.* together... etc).
+      #
+      # Default: ["([^.])+"]
```
michaelpj, line 372: great doc!

lib/Language/Haskell/Stylish/Step/Imports.hs
```diff
@@ -103,6 +119,27 @@ data LongListAlign
     | Multiline -- multiline
     deriving (Eq, Show)
 
+data Pattern = Pattern
```
michaelpj, line 122: Haddock?

lib/Language/Haskell/Stylish/Step/Imports.hs
```diff
@@ -103,6 +119,27 @@ data LongListAlign
...
+  { regex  :: Regex
```
michaelpj, line 123: it's fairly clear from reading the code, but perhaps worth a comment that this is a cached computed field derived from `string`.

lib/Language/Haskell/Stylish/Step/Imports.hs
```diff
@@ -103,6 +119,27 @@ data LongListAlign
...
+  , string :: String
+  }
+
+instance Show Pattern where show = show . string
+
+instance Eq Pattern where (==) = (==) `on` string
+
+instance IsString Pattern where
+  fromString = either error id . parsePattern
```
michaelpj, line 132: The `error` seems a bit risky. Are we sure that's not going to sneak in somewhere unexpected? Would it be much worse to just have `unsafeParsePattern`?

lib/Language/Haskell/Stylish/Step/Imports.hs
```diff
@@ -103,6 +119,27 @@ data LongListAlign
...
+
+instance A.FromJSON Pattern where
+  parseJSON = \case
+    A.String string -> fromRight mzero (pure <$> parsePattern (T.unpack string))
```
michaelpj, line 136: use `withText`? and `fail` with the error if parsing fails?

lib/Language/Haskell/Stylish/Step/Imports.hs
```diff
@@ -160,6 +201,63 @@ formatImports maxCols options moduleStats rawGroup =
...
+--------------------------------------------------------------------------------
+groupAndFormat
+  :: Maybe Int
+  -> Options
+  -> ImportStats
+  -> [NonEmpty (GHC.LImportDecl GHC.GhcPs)]
+  -> Editor.Edits
+groupAndFormat maxCols options moduleStats groups = edits
+  where edits = Editor.changeLines block (const regroupedLines)
+        regroupedLines =
```
michaelpj, line 213: as a reader, I do find it very helpful for these sorts of bindings to have type signatures...

lib/Language/Haskell/Stylish/Step/Imports.hs
```diff
@@ -160,6 +201,63 @@ formatImports maxCols options moduleStats rawGroup =
...
+          intercalate [""] $
+          map (formatImports maxCols options moduleStats) grouped
+
+        grouped = groupByPatterns (groupPatterns options) imports
+        imports = concatMap toList groups
+
+        block = Block
+          (GHC.srcSpanStartLine . src $ head imports)
```
michaelpj, line 221: I think this is not safe in the case where `groups` is empty. Which I think might be possible if you format a file with no imports (i.e. just using the `Prelude`).

lib/Language/Haskell/Stylish/Step/Imports.hs
```diff
@@ -160,6 +201,63 @@ formatImports maxCols options moduleStats rawGroup =
...
+          GHC.srcSpanToRealSrcSpan . GHC.getLocA
+
+groupByPatterns
+  :: [Pattern]
+  -> [GHC.LImportDecl GHC.GhcPs]
+  -> [NonEmpty (GHC.LImportDecl GHC.GhcPs)]
+groupByPatterns patterns allImports = go patterns allImports []
+  where
+    go [] [] groups            = groups
+    go [] imports groups       = groups <> [NonEmpty.fromList imports]
```
michaelpj, line 233: Almost certainly doesn't matter given the sizes of lists we're looking at, but this function does a lot of concatenating things on the back of lists, which is linear. I'd either use a `DList` or build it in reverse order and reverse at the end.

lib/Language/Haskell/Stylish/Step/Imports.hs
```diff
@@ -160,6 +201,63 @@ formatImports maxCols options moduleStats rawGroup =
...
+        go ps rest (groups <> groups')
+
+    extract p imports =
+      let
+        (matched, rest) =
+          partition (matches p) imports
+        subgroups =
+          groupBy ((==) `on` capture p) $ sortOn (capture p) matched
+      in
+        (map NonEmpty.fromList subgroups, rest)
```
michaelpj, line 247: `NonEmpty.fromList` is partial! Are we sure there won't be any empty lists in there? `mapMaybe NonEmpty.nonEmpty` is I think the same but safe.

lib/Language/Haskell/Stylish/Step/Imports.hs
```diff
@@ -160,6 +201,63 @@ formatImports maxCols options moduleStats rawGroup =
...
+        (groups', rest) = extract p imports
+      in
+        go ps rest (groups <> groups')
+
+    extract p imports =
+      let
+        (matched, rest) =
+          partition (matches p) imports
+        subgroups =
+          groupBy ((==) `on` capture p) $ sortOn (capture p) matched
```
michaelpj, line 245: It looks like we're going to run `p` on the same text 3 times doing this. But regex matching is fast so probably not worth worrying about.

lib/Language/Haskell/Stylish/Step/Imports.hs
```diff
@@ -160,6 +201,63 @@ formatImports maxCols options moduleStats rawGroup =
...
+        (map NonEmpty.fromList subgroups, rest)
+
+    matches :: Pattern -> GHC.LImportDecl GHC.GhcPs -> Bool
+    matches Pattern { regex } import_ = match regex $ moduleName import_
+
+    capture :: Pattern -> GHC.LImportDecl GHC.GhcPs -> String
+    capture Pattern { regex } import_ =
+      case getAllTextSubmatches $ match regex $ moduleName import_ of
+        (_ : s : _) -> map toLower s -- lowercase for case-insensitive sorting
+        _           -> ""            -- all grouped together
```
michaelpj, line 256:
```suggestion
        _           -> ""            -- constant grouping key, so everything will be grouped together
```

lib/Language/Haskell/Stylish/Step/Imports.hs
```diff
@@ -160,6 +201,63 @@ formatImports maxCols options moduleStats rawGroup =
...
+      in
+        (map NonEmpty.fromList subgroups, rest)
+
+    matches :: Pattern -> GHC.LImportDecl GHC.GhcPs -> Bool
+    matches Pattern { regex } import_ = match regex $ moduleName import_
+
+    capture :: Pattern -> GHC.LImportDecl GHC.GhcPs -> String
+    capture Pattern { regex } import_ =
+      case getAllTextSubmatches $ match regex $ moduleName import_ of
+        (_ : s : _) -> map toLower s -- lowercase for case-insensitive sorting
```
michaelpj, line 255: Well, that answers my earlier question: subsequent matching groups are ignored. I think this should throw an error if there's more than one, better to tell the user they did something unexpected.

tests/Language/Haskell/Stylish/Step/Imports/Tests.hs
```diff
@@ -930,3 +1009,248 @@ case38 = assertSnippet (step (Just 80) $ fromImportAlign File)
...
+  [ "import Something.A"
+  , "import SomethingElse.A"
+  , "import SomeThing.B"
+  , "import SomeThingelse.B"
+  ]
+  [ "import           Something.A"
+  , "import           SomeThing.B"
+  , ""
+  , "import           SomethingElse.A"
+  , "import           SomeThingelse.B"
```
michaelpj, line 1025: You know, I don't know how I feel about this. Module names are case-sensitive in Haskell, so why would I group these together?

tests/Language/Haskell/Stylish/Step/Imports/Tests.hs
```diff
@@ -930,3 +1009,248 @@ case38 = assertSnippet (step (Just 80) $ fromImportAlign File)
...
+    , "import ProJect.WrongCapitalization"
+    ]
+    [ "import Project"
+    , "import Project.Something"
+    , ""
+    , "import Control.Monad"
+    , "import qualified Data.Acid as Acid"
+    , "import Data.Default.Class (Default (def))"
+    , "import qualified Herp.Derp.Internal.Types.Foobar as Internal (bar, foo)"
+    , "import ProJect.WrongCapitalization"
```
michaelpj, line 1250: Wait, *now* we're case-sensitive? That seems inconsistent.

michaelpj (approved):
Very nice code and tests. Mostly just nitpicking comments.

lib/Language/Haskell/Stylish/Step/Imports.hs
TikhonJelvis: Happy to add haddocks here—I didn't write any because the rest of the codebase didn't seem to use haddock comments.

data/stylish-haskell.yaml
TikhonJelvis: Great call. I'll add a bit more explanation in the docs—ideally, I want to provide sufficient explanation and examples so that people could at least configure rules based on specific module name components without prior familiarity with regular expression syntax.

michaelpj: I think _most_ users will want to have a leading `^`, since I expect "top-down" grouping to be the most common desire.

lib/Language/Haskell/Stylish/Step/Imports.hs
michaelpj: "Be the change you want to see in the world" or something? :)

data/stylish-haskell.yaml
TikhonJelvis: thanks :)

lib/Language/Haskell/Stylish/Step/Imports.hs
TikhonJelvis: In this case `subgroups` can't be empty by construction. Not sure it would be worth handling the empty case if it required several extra lines of code, but the `mapMaybe NonEmpty.nonEmpty` version is a great way to avoid the problem without that.

michaelpj: Why can't `subgroups` be empty? What if `matched` is empty? Which it could be if you have a regex that doesn't match any of the imports?

lib/Language/Haskell/Stylish/Step/Imports.hs
TikhonJelvis: Yep, exactly my reasoning. I started out trying to write a version that only ran the pattern once, but it ended up making the logic quite a bit more complex, and I decided that it wasn't worth it. Matching with regex-tdfa should be fast and I expect module names would break GHC before getting long enough to cause performance problems here.

michaelpj: Perhaps worth a short comment pointing that out.

lib/Language/Haskell/Stylish/Step/Imports.hs
TikhonJelvis: Sorry, I misspoke: `subgroups` can be empty, but it can't contain empty lists because `groupBy` will never generate empty lists. The `NonEmpty.fromList` call applies to the lists inside `subgroups`, not `subgroups` itself.

TikhonJelvis: In an ideal world we'd have `List.groupBy :: (a -> a -> Bool) -> [a] -> [NonEmpty a]`. But I also understand why we don't: converting from `NonEmpty a` to `[a]` is annoying, especially since it's logically a no-op. That's one of the big practical limitations I've found with H-M type systems: talking about *restrictions* on types adds a lot of distracting boilerplate which is not really necessary.

A refinement type system like LiquidHaskell avoids this problem and it would be pretty sweet to have that easily integrated into GHC just to reduce the cost of using things like `NonEmpty`. As-is, it's only worth using LiquidHaskell for projects that really care about proving additional properties about the code, but it requires too much setup just for convenience or a bit of additional type expressiveness.

lib/Language/Haskell/Stylish/Step/Imports.hs
TikhonJelvis: Great call. Avoiding quadratic things seems like the way to go even if it doesn't matter now—doesn't take *that much* to turn it into a real problem.

lib/Language/Haskell/Stylish/Step/Imports.hs
TikhonJelvis: I tend to use partial functions in my own `IsString` instances because I figure that mistakes in string literals are very unlikely to make it past even cursory tests. Hasn't bitten me yet, but maybe I'm underestimating the problem :).

In this case, the instance is only used once and `Pattern` isn't exported from this module. That could change if we end up using `Pattern` in other parts of the code later though, so making it explicit makes sense. For now, I can inline the `either error id` bit into the one place where I used the `IsString` instance previously.

TikhonJelvis: Oh wait, never mind, I'll probably want to export the unsafe version for use in tests too.

michaelpj: I've seen people use `fromString` directly, though. Not a huge deal.

lib/Language/Haskell/Stylish/Step/Imports.hs
TikhonJelvis: Just realized that regex-tdfa does not support non-capturing groups at all, which makes the whole capturing thing pretty annoying—there's no way to group stuff without triggering the sub-grouping logic. That could be reason enough to move this to move to [regex-pcre][1] instead. Then using multiple capture groups could result in an error with a message that recommended using non-capturing groups instead.

However, looks like regex-pcre has [an open issue with Unicode support][2], so that isn't great either :/

I'll push some commits for your other suggestions, but going to need a bit more time to think through this capture group stuff.

[1]: https://hackage.haskell.org/package/regex-pcre

[2]: https://github.com/haskell-hvr/regex-pcre/issues/3

tests/Language/Haskell/Stylish/Step/Imports/Tests.hs
TikhonJelvis: Yeah, not sure. I saw that the import logic for sorting was case-insensitive, so I wanted the grouping logic to match that. Might be better not to do that at all though.

tests/Language/Haskell/Stylish/Step/Imports/Tests.hs
TikhonJelvis: Yeah, that's a bit odd. To make things consistent, I made all of the subgrouping logic case sensitive.

TikhonJelvis:
@michaelpj Great suggestions, thanks! I made changes for everything except the capture group stuff, which I'll have to think about a bit more—not supporting non-capturing groups is a bigger headache than I had thought originally, since a lot of reasonable regular expressions can't be written at all without grouping.

michaelpj:
Two random suggestions on capture groups:
- Add another option for whether to do sub-grouping on capture groups or not. Then you can choose whether to write nice regexes with groups and not worry about your imports getting grouped weirdly; or write annoying regexes with the nice grouping behaviour you want.
- Add more structure to the config for patterns, so we have something like:
```
- group_conditions:
    - pattern:
        - regex: `(hello)(world)`
        - group_on_subgroup: 2
    - pattern:
        - regex: `whatever`
```

TikhonJelvis:
@michaelpj Thanks! Those suggestions really helped—was a bit stuck on the design otherwise. Having a structured setting for each pattern seems both more usable and more flexible for the future.

If we end up having two settings anyhow, may as well allow separate patterns as opposed to sub-groups. Since the last time I've looked at this PR, I've found even *more* ways in which capture groups are unintuitive: if a group matches multiple times (`(...)*` for example), do we expect to capture the *first* instance of `(...)`, the *last* instance or all of them? I saw a Twitter poll about this and it was clear that there was no "obvious" consensus. `regex-tdfa` always returns the *last* sub-match, which seems inconvenient for dealing with module names.

So now I'm thinking about something like this, with `match` required and `group_by` optional.

``` yaml
- group_conditions:
  - match: regex
    group_by: regex
  - match: regex
```

michaelpj:
That design seems even better - I agree that capture groups look increasingly wonky the more you think about it!

TikhonJelvis:
Thanks! Just implemented it—didn't take too many changes from the previous version of the code.

data/stylish-haskell.yaml
```diff
@@ -294,6 +294,97 @@ steps:
...
+      #    - match: "^MyApp\\>"
+      #
+      #  > import MyApp
+      #  > import MyApp.Foo
+      #  >
+      #  > import Control.Monad
+      #  > import MyApps
+      #  > import Test.MyApp
+      #
+      # A rule can also optionally have a sub_group pattern. Import
```
jaspervdj, line 351:
```suggestion
      # A rule can also optionally have a sub_group pattern. Imports
```

data/stylish-haskell.yaml
```diff
@@ -294,6 +294,97 @@ steps:
...
+      #   > import Control.Monad.MonadError
+      #
+      # Default: false
+      group_imports: false
+
+      # A list of rules specifying how to group modules and how to
+      # order the groups.
+      #
+      # Each rule has a match field; the rule only applies to module
+      # names matched by this pattern.
```
jaspervdj, line 331:
```suggestion
      # names matched by this regex pattern.
```

data/stylish-haskell.yaml
```diff
@@ -294,6 +294,97 @@ steps:
...
+      # Rules are processed in order, so only the *first* rule that
+      # matches a specific module will apply. Any module names that do
+      # not match a single rule will be put into a single group at the
+      # end of the import block.
+      #
+      # Example: group MyApp modules first, with everything else in
+      # one group at the end.
+      #
+      #  group_rules:
+      #    - match: "^MyApp\\>"
```
jaspervdj, line 342:
```suggestion
      #    - match: "^MyApp\\."
```

Typo?

jaspervdj (commented):
Some minor comments about the docs, this is looking great!

data/stylish-haskell.yaml
TikhonJelvis: Oh, `\\>` is an anchor for the end of the word. `^MyApp\\>` would match the module name `MyApp` while `^MyApp\\.` wouldn't.

I had a note about that in the documentation at one point, but looks like I lost it in one of the revisions. I'll add a note about it after the bit about ^ and $, since it's a handy feature from POSIX regexps that is different in PCRE (which uses `\b` instead) and that people might not know about.

jaspervdj: Awesome, yeah, I only knew about `\b`!

TikhonJelvis:
I've started using this on a toy project and I'm happy with it—between this and HLS, I can have neatly organized imports without having to think about it any more :)

TikhonJelvis:
@jaspervdj Is this ready to merge? I've been using it on one of my personal projects for a couple months now without problems. After getting it working locally for myself, I forgot it wasn't in the normal version of the tool yet :P

TikhonJelvis:
Okay, should be squashed into a single commit unless I messed something up with git :)

jaspervdj:
Thanks a lot!

TikhonJelvis:
@jaspervdj Would you mind cutting a release with this feature? I've figured out a clean way to use a local build of `stylish-haskell` for my own projects so there's no hurry for me, but it would make things cleaner and let other people use it conveniently.

Thanks!

jaspervdj:
Released in `v0.14.3.0`, sorry for the delay!

TikhonJelvis:
Sweet, thanks!

PR #402, opened 2022-04-27
Add ghc-lib flag to force use ghc-lib-parser even if ghc is supported

July541:
Hello, this pr adds a ghc-lib flag to force use ghc-lib-parser even if ghc is supported. HLS will benefit a lot since we can support hls-stylish-haskell-plugin and hls-hlint-plugin on ghc9.2 simultaneously.  I would appreciate it if this pr is accepted with a new release. Thanks in advance.

The necessity is here https://github.com/haskell/haskell-language-server/pull/2854#discussion_r858874680

### DIFF
--- a/stylish-haskell.cabal
+++ b/stylish-haskell.cabal
@@ -1,6 +1,6 @@
 Cabal-version: 2.4
 Name:          stylish-haskell
-Version:       0.14.2.0
+Version:       0.14.1.0
 Synopsis:      Haskell code prettifier
 Homepage:      https://github.com/haskell/stylish-haskell
 License:       BSD-3-Clause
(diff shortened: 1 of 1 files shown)


stylish-haskell.cabal
```diff
@@ -1,6 +1,6 @@
 Cabal-version: 2.4
 Name:          stylish-haskell
-Version:       0.14.1.0
+Version:       0.14.2.0
```
jaspervdj, line 3:
```suggestion
Version:       0.14.1.0
```

jaspervdj (approved):
This is excellent, I was looking into supporting both ghc9 and ghc9.2 but the API differences turned out to be more than I expected.  Thanks a lot!  I'm happy to release this today.

Just one minor nitpick: can we remove the version bump?  I prefer to do this in a separate commit since I have a bunch of scripts for this.

July541:
FYI, I have a branch on ghc9.0 https://github.com/July541/stylish-haskell/tree/ghc-9.0

PR #417, opened 2022-10-17
Move to nix flakes

lukasz-golebiewski:


lukasz-golebiewski:
Should wait for https://github.com/haskell/stylish-haskell/pull/419

PR #419, opened 2022-10-18
Bump resolver

lukasz-golebiewski:
The resolver version is bumped to nightly-2022-06-06 which is the latest with ghc 9.2.2

This is done so that ghc-lib-parser* dependencies do not need to be defined as extra-deps


lukasz-golebiewski:
Should wait for https://github.com/haskell/stylish-haskell/pull/417

jaspervdj (approved):

PR #421, opened 2022-10-22
Support ghc-9.4

July541:


July541:
It is only buildable with ghc-lib, build with ghc-9.4 requires to specify flag for ghc-lib-parser-ex...

July541:
Slightly ping @jaspervdj, any update required here?

jaspervdj:
Looks good to me, thanks a lot for taking this on!

jaspervdj (approved):

michaelpj:
It would be great to get a release with this in, I think that's all that's stopping us supporting stylish-haskell for 9.4 in HLS.

EncodePanda:
@jaspervdj I think we can cut a release, correct?

PR #424, opened 2022-11-06
Deriving alignment for enums

### LINKED ISSUE #411: `deriving` alignment for enums
Is there a way to align `deriving` close at the next line? With default options (if we uncomment format records rules formatter change
```hs
data Tmp1 = A | B | C
  deriving (Show, Eq)
```
to
```hs
data Tmp1 = A | B | C deriving (Eq, Show)
```
This becomes more ugly with large enums:
```hs
data DefType = DefTEnv | DefTPref | DefTCmd | DefTInl | DefTMode | DefTSort deriving
    ( Eq
    , Ord
    , Show
    )
```
instead of
```hs
data DefType = DefTEnv | DefTPref | DefTCmd | DefTInl | DefTMode | DefTSort 
  deriving (Eq, Ord, Show)
```
Maybe I'm doing something wrong, but I cannot see an option to manage this behavior

Lev135:
Fixes #411

I've changed printer's behavior for alignment of deriving closes: now it adds newline and indentation before `deriving` even if `onelineEnum` is enabled:
```hs
data Foo = A | B | C
  deriving (Eq, Show)
```
instead of
```hs
data Foo = A | B | C deriving (Eq, Show)
```
This behavior is what is said in docs:

> Deriving clauses are always on separate lines

Surely we can add an option to preserve the previous behavior, but I don't think that it's desired for several reasons:

- it conflicts with docs
- it leads to an awful output when enum is large, that even not highlight well:
```hs
data DefType = DefTEnv | DefTPref | DefTCmd | DefTInl | DefTMode | DefTSort deriving
    ( Eq
    , Ord
    , Show
    )
```

However, if you think otherwise, I can add an option. Maybe while `break_enums` option was discussed the reasons of this choice were expained, but I couldn't find anything.

Only two tests were affected by this change:

- case 20 (establish the `break_enums` option) 

Before:
```hs
data Tag = Title | Text deriving (Eq, Show)
```
After:
```hs
data Tag = Title | Text 
  deriving (Eq, Show)
```

- case 55 (issue #262, not related to `deriving` problem)

Before:
```hs
data Foo = Foo deriving (Z, Y, X, Bar, Abcd)
```
After:
```hs
data Foo = Foo 
  deriving (Z, Y, X, Bar, Abcd)
```

### DIFF
--- a/lib/Language/Haskell/Stylish/Step/Data.hs
+++ b/lib/Language/Haskell/Stylish/Step/Data.hs
@@ -196,8 +196,9 @@ putDataDecl cfg@Config {..} decl = do
     let derivingComments = deepAnnComments (GHC.dd_derivs defn)
 
     when (hasDeriving decl) do
-        if onelineEnum && null derivingComments then
-            space
+        if onelineEnum && null derivingComments then do
+            newline
+            spaces cDeriving
         else do
             forM_ derivingComments $ \lc -> do
                 newline
--- a/tests/Language/Haskell/Stylish/Step/Data/Tests.hs
+++ b/tests/Language/Haskell/Stylish/Step/Data/Tests.hs
@@ -1377,7 +1377,7 @@ case66 :: Assertion
 case66 = assertSnippet (step indentIndentStyle) input input
   where
     input =
-      [ "data Tmp1 = A | B | C"
+      [ "data Foo = A | B | C"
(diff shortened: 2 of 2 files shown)


Lev135:
Oh, I've found related issue: #379. I think it should be closed by this pull request

Issue #409, opened 2022-08-30
feature: marker to ignore the following block


imuli:
I occasionally run into situations where I want to manually format a block, e.g. with `servant-openapi3`:

```haskell
-- | The Work API
-- ...
data WorkAPI n mode = WorkAPI
   { workCreate :: mode
      :- Summary "Create a work from a design."
      :> Description "..."
      :> NamedRoutes (RunAPI (Design n) (WorkId n))
   , workIdDesignGet :: mode
      :- Summary "Fetch the most recent design on a work."
      :> Description "..."
      :> WorkId n :> "design" :> NamedRoutes (GetAPI (Design n))
   }
  deriving (Generic)
```

`stylish-haskell`, of course, will reformat that so each type definition is on a single line, which is what I want in most cases. This would also provide a workaround while waiting on a bug fix (or for me in the case of record GADTs with docs, bug filing!).

- `{- STYLISH ignore -}` typically wouldn't get special syntax highlighting.
- `{-# STYLISH ignore #-}` gets special syntax highlighting, but also makes ghc noisy without `-Wno-unrecognized-pragmas`
- `{-# ANN myDef ("STYLISH ignore" :: String) #-}` gets special syntax highlighting, but also requires attaching to a specific definition (e.g. not usable for imports).

andys8:
Would help with the `haskell-language-server` project where `stylish-haskell` is in use but fails to parse some files. Skipping these sections would format the rest of the file and help with [failing `pre-commit` tasks](https://github.com/haskell/haskell-language-server/actions/runs/3077951377/jobs/4973176369#step:31:87).

```
plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs: <string>:442:26: error: parse error on input `='
ghcide/src/Development/IDE/Import/FindImports.hs: <string>:127:5: error:
    parse error (possibly incorrect indentation or mismatched brackets)
plugins/hls-refactor-plugin/src/Development/IDE/GHC/Compat/ExactPrint.hs: <string>:4:1: error: parse error on input `module'
ghcide/src/Development/IDE/Core/Compile.hs: <string>:372:7: error: parse error on input `go'
```

imuli:
This https://github.com/imuli/stylish-haskell/commit/22fd02d515394b3509e687f1e23b767399a4c07c commit in my personal version is the minimal thing that meets (some) of my needs. It only handles the comment inside the data declaration though:

```haskell
[...]

PR #438, opened 2022-12-29
Disabling formatting of the block of code

Lev135:
Here is one of the possible solutions to this issue. It's very straightforward: just collecting all edits and then filtering them and applying only those, that doesn't intersect with the block of code between `{- stylish_disable -}` and `{- stylish_enable -}`.

Filtering is applied by rows only, so, in this sample
```hs
data Foo = Foo Int {- stylish_disable -}
{- stylish_enable -}
```
data will not be reformated. In the first my iplementation columns also were taken into account, but this made the code much more havier and didn't make much sense for me: such constructions as in example above seems too artificial.

Anyway this issue was discussed many times here (#409 is the latest one). So it seems to be very helpful feature, not for me only. So it would be nice to discuss and resolve it one way or another

### DIFF
--- a/lib/Language/Haskell/Stylish/Editor.hs
+++ b/lib/Language/Haskell/Stylish/Editor.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE BlockArguments #-}
 
 --------------------------------------------------------------------------------
 -- | This module provides you with a line-based editor. It's main feature is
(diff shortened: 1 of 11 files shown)


jaspervdj (commented):
I like the approach of doing this in Editor — I will look at the implementation in more detail once I’m back from holidays.

Lev135:
@jaspervdj
Sorry to disturb you, but I'm wonder, whether any progress with merging PRs can be expected in the nearest future. The reason is that I'm working on my project and the lack of this feature and some bugs in stylish (mostly #425) cause many inconveniences. Of course, I can build this all in my fork and use it instead of the main repo, but it would be much clearer if it can be used directly. Please, let me note if I can somehow facilitate the process.
Thank you

Issue #414, opened 2022-09-28
parse error on input on "type (*)"


Mikolaj:
Stylish-haskell works great with type-level (*) taken from

```
import GHC.TypeLits
```

but it chokes at

```
import GHC.TypeLits (type (*))
```

with

```
Warning (stylish-haskell): stylish-haskell failed: <string>:16:48: error:
    parse error on input `*'
```

Lev135:
Is it still an issue? I tried to reproduce it with `stylish-haskell 0.14.3.0` and `ghc-9.2.4`: with `TypeOperators` and `NoStarIsType` stylish works fine, without `NoStarIsType` both ghci and stylish produce the same parse error. Maybe this was a problem only before `NoStarIsType` was fixed? Or I'm just looking on incorrect combination of settings

Mikolaj:
I just tried to verify by compiling stylish-haskell-0.14.3.0 with ghcup-installed ghc-9.2.5 and got the following:

```
[ 4 of 26] Compiling Language.Haskell.Stylish.GHC ( lib/Language/Haskell/Stylish/GHC.hs, dist/build/Language/Haskell/Stylish/GHC.o, dist/build/Language/Haskell/Stylish/GHC.dyn_o )

lib/Language/Haskell/Stylish/GHC.hs:71:32: error:
    • Couldn't match expected type ‘GHC.Settings’
                  with actual type ‘ghc-lib-parser-9.2.5.20221107:GHC.Settings.Settings’
      NB: ‘GHC.Settings’
            is defined in ‘GHC.Settings’ in package ‘ghc-9.2.5’
          ‘ghc-lib-parser-9.2.5.20221107:GHC.Settings.Settings’
            is defined in ‘GHC.Settings’
                in package ‘ghc-lib-parser-9.2.5.20221107’
    • In the first argument of ‘defaultDynFlags’, namely
        ‘GHCEx.fakeSettings’
      In the expression:
        defaultDynFlags GHCEx.fakeSettings GHCEx.fakeLlvmConfig
      In an equation for ‘baseDynFlags’:
          baseDynFlags
            = defaultDynFlags GHCEx.fakeSettings GHCEx.fakeLlvmConfig
   |
71 | baseDynFlags = defaultDynFlags GHCEx.fakeSettings GHCEx.fakeLlvmConfig
   |                                ^^^^^^^^^^^^^^^^^^

lib/Language/Haskell/Stylish/GHC.hs:71:51: error:
[...]

PR #439, opened 2022-12-29
Fixed NoXyz issue for .cabal files

Lev135:
This is just like #412, but for .cabal files. This is what actually caused #414, so I hope it will be closed by the PR.

Also I've improved a bit tests for config parsing. I think they are not optimal, but now they are a bit better, than they were.

For the future plans I think it would be nice to some algebraic type for dealing with extensions. The current solution "via show" seems ugly and error-prone.


jaspervdj (approved):

jaspervdj:
Thank you!

PR #440, opened 2023-01-03
Disapearing of `DEPRECATED` pragma attached to module fixed

### LINKED ISSUE #433: DEPRECATED pragma is removed by module_header step
[The DEPRECATED pragma](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/pragmas.html?highlight=deprecated#pragma-DEPRECATED) is removed by module_header step.

Example:

```haskell
module X {-# DEPRECATED "Do not use this" #-}
    ( foo
    ) where

foo :: String
foo = "foo"
```

stylish-haskell just removes the pragma.

Lev135:
Fixes #433


jaspervdj:
Thanks!

PR #448, opened 2023-03-05
Cabal GitHub action for GHC 9.0, 9.2, 9.4

peterbecich:
separate from the current Stack GitHub Action

Build with GHC 9.2 looks good, maybe https://github.com/haskell/stylish-haskell/issues/405 can be closed

The build with GHC 9.4 fails: https://github.com/haskell/stylish-haskell/issues/449

### DIFF
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -13,10 +13,6 @@ jobs:
       fail-fast: false
 
     steps:
-    - name: Get the version
-      id: get_version
-      run: 'echo ::set-output name=version::${GITHUB_REF#refs/tags/}'
-
     - uses: actions/checkout@v2
 
     - uses: haskell/actions/setup@v2
@@ -27,7 +23,7 @@ jobs:
     - uses: actions/cache@v2
       name: Cache ~/.cabal
       with:
-        path: ~/.stack
+        path: ~/.cabal
         key: "${{ runner.os }}-${{ matrix.ghc }}-v9-${{ hashFiles('stylish-haskell.cabal') }}"
 
     - name: Build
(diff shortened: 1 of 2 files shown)


.github/workflows/cabal.yml
```diff
@@ -0,0 +1,51 @@
...
+    name: Build on ${{ matrix.os }} GHC ${{ matrix.ghc }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macOS-latest]
+        ghc: ["9.0.2", "9.2.7", "9.4.4"]
+      fail-fast: false
+
+    steps:
+    - name: Get the version
```
jaspervdj, line 16: We probably don’t need this step

.github/workflows/cabal.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    - uses: haskell/actions/setup@v2
+      name: Setup Haskell Cabal
+      with:
+        ghc-version: ${{ matrix.ghc }}
+
+    - uses: actions/cache@v2
+      name: Cache ~/.cabal
+      with:
+        path: ~/.stack
```
jaspervdj, line 30: Path should likely be updated?

.github/workflows/cabal.yml
```diff
@@ -0,0 +1,51 @@
...
+        key: "${{ runner.os }}-${{ matrix.ghc }}-v9-${{ hashFiles('stylish-haskell.cabal') }}"
+
+    - name: Build
+      run: cabal build --enable-tests
+      id: build
+
+    - name: Test
+      run: cabal test
+
+    - name: Build artifact
```
jaspervdj, line 40: Everything below this is not needed I think

jaspervdj (changes requested):
Thanks for the PR!

.github/workflows/cabal.yml
peterbecich: you're right, thanks

jaspervdj (approved):

jaspervdj:
Looking good now, thanks a lot!

peterbecich:
Thank you, @jaspervdj

PR #455, opened 2023-05-26
Add option to remove pragmas implied by the language variant

mrBliss:
For example, by setting `language_variant` to `GHC2021`, all the extensions implied by the GHC2021 language variant will be removed as redundant language pragmas.

### DIFF
--- a/data/stylish-haskell.yaml
+++ b/data/stylish-haskell.yaml
@@ -426,6 +426,19 @@ steps:
       # is set to true, it will remove those redundant pragmas. Default: true.
       remove_redundant: true
 
+      # When remove_redundant is enabled, extensions that are implied by the
+      # chosen language variant will also be removed. The following language
+      # variants are supported:
+      #
+      # - GHC2021
+      #
+      # - Haskell2010
+      #
+      # - Haskell98
+      #
+      # Default: Haskell2010
+      language_variant: Haskell2010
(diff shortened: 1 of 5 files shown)


data/stylish-haskell.yaml
```diff
@@ -426,6 +426,19 @@ steps:
...
+      # variants are supported:
+      #
+      # - GHC2021
+      #
+      # - Haskell2010
+      #
+      # - Haskell98
+      #
+      # Default: Haskell2010
+      language_variant: Haskell2010
```
mrBliss, line 440: 1. I was considering whether there should be a way to retain the former behaviour of `remove_redundant: true`, i.e., just remove `BangPatterns` and `ViewPatterns` when unused in the module. I decided not to add it because I couldn't think of a good way to make it fit the `language_variant` option (which language variant would that option be?) and the usefulness of the former behaviour is rather limited, IMHO.

2. What about projects that are mix of different language variants? The mix would be across components (library vs. testsuite, etc.)

michaelpj:
This would be great to have! Could we also parse this setting out of the `.cabal` file like we do for `default-extensions`? I don't think that would directly solve your problem about different components with different languages (e.g. for default extensions it seems that we just take the union of all default extensions across all components!), but it's a step in the right direction.

PR #459, opened 2023-06-23
Update to support GHC 9.6

### LINKED ISSUE #458: GHC 9.6 support
It's been out for a little while. I was able to get the dependencies sorted and fix a couple of compile errors, but there are more. WIP branch [here](https://github.com/michaelpj/stylish-haskell/tree/mpj/9.6-support) if anyone else feels motivated, otherwise perhaps I will get back to it.

michaelpj:
Fixes #458

### DIFF
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -27,7 +27,7 @@ jobs:
         key: "${{ runner.os }}-${{ matrix.ghc }}-v9-${{ hashFiles('stylish-haskell.cabal') }}"
 
     - name: Build
-      run: cabal build --enable-tests
+      run: cabal build 
       id: build
 
     - name: Test
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest]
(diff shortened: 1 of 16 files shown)


michaelpj:
Hmm, looks like the latest `ghc-lib-parser` doesn't support 9.0. I'm not sure what to do about that: AFAICT that's their intended support policy, and to do anything else we'd need to add a bunch of CPP, which is exactly what we're avoiding with `ghc-lib-parser`. So maybe it's just time to drop 9.0?

michaelpj:
I added 9.6 to the workflow and removed 9.0, but it probably won't run since my PR is from a fork.

jaspervdj (approved):
Looks good!  I was about to bump `optparse-applicative` in another PR, thanks for taking care of that as well!

michaelpj:
Yeah, I'm just trying to make stack happy...

michaelpj:
Okay, it's all green!

jaspervdj:
Thanks a lot!

michaelpj:
It would be great to get a release, then we can get this supported in HLS.

jaspervdj:
Released as v0.14.5.0

michaelpj:
Thanks, I've made a PR to enable it in HLS :)

PR #468, opened 2023-11-07
Add MultilineOnly LostListAlign variant

sorki:
Adding `MultlineOnly` mode which forces multiline output without trying inline first. Related to #467


sorki:
cc @jonathanknowles

PR #469, opened 2023-11-07
Add EntityListAlign to imports step

sorki:
This patch add support for alignment of lists of constructors, class methods, fields names in import lists entities.

`ELMultiline` causes

```
import Foo (Bar(x, y))
```

To be expanded to

```
import Foo
  (Bar
    ( x
    , y
    )
  )
```

This is useful in combination with `LongListAlign` `Multiline` to get a git friendly formatting.

### DIFF
--- a/data/stylish-haskell.yaml
+++ b/data/stylish-haskell.yaml
@@ -218,6 +218,29 @@ steps:
       # Default: inline
       long_list_align: inline
 
+      # Entity list align can be used to alter alignment of import list entities
+      # like 'Bar' record names in the following example:
+      #
+      #   > import Foo (Bar(x, y))
+      #
+      # - inline: Entities are always formatted inline
+      #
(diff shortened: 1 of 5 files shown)


sorki:
A less contrived (and prettier) test requires #468 for forcing multiline only output, which causes the output to look like:

```haskell
import Linear
  ( V2(V2)
  , V3(V3)
  , V4(V4)
  )

import Prelude
  ( Bool(True)
  , Eq((==))
  , Num 
      ( (*) 
      , (+) 
      )   
  )
```

The special case of single entity is needed to prevent
```haskell
import Linear
  ( V2
      ( V2
      )
  , V3
      ( V3
      )
  , V4
      ( V4
      )
  )
```

This in combination with #468 allows for the most git-friendly formatting when explicit import lists are used (typically when `-Wmissing-import-lists` is enabled).

PR #471, opened 2023-12-15
9.8 support

michaelpj:
Work-in-progress. I did the obvious changes. There are a few test failures in the language pragma tests and I don't know why, otherwise it seems to work.

### DIFF
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest]
-        ghc: ["9.2", "9.4", "9.6", "9.8"]
+        ghc: ["9.4", "9.6", "9.8"]
       fail-fast: false
 
     steps:
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest]
-        ghc: ["9.2", "9.4", "9.6"]
+        ghc: ["9.2", "9.4", "9.6", "9.8"]
       fail-fast: false
 
     steps:
--- a/lib/Language/Haskell/Stylish/Printer.hs
+++ b/lib/Language/Haskell/Stylish/Printer.hs
@@ -239,7 +239,7 @@ putType ltp = case GHC.unLoc ltp of
     putOutputable ltp
   GHC.HsQualTy {} ->
     putOutputable ltp
-  GHC.HsAppKindTy _ _ _ ->
+  GHC.HsAppKindTy _ _ _ _ ->
     putOutputable ltp
   GHC.HsListTy _ _ ->
     putOutputable ltp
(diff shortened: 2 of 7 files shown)


michaelpj:
Ah, so taking the new `ghc-lib-parser` will force us to drop 9.2. I don't know if that's okay.

michaelpj:
Okay, I'm fairly confused, I've traced this back to the exactprint annotations for the module not containing the `ViewPatterns` pragma, which is baffling.

michaelpj:
I tracked it down. It seems that if you don't provide a `module X where` header, then something goes wrong with the attaching of the exactprint annotations (`moduleLanguagePragmas` does not produce the right comments, and printing out the annotations confirms that). What seems to happen is:
```
{-# LANGUAGE P1 #-} 
{-# LANGUAGE P2 #-}
x = 1
```
then we see only the P1 comment attached to the module. Notably, this other case works:
```
{-# LANGUAGE P1 #-} 
{-# LANGUAGE P2 #-} 
```

My suspicion is that something has changed in the parser so maybe `pragma1; pragma2; binding` now parses as `pragma1; module; pragma2; binding` rather than `pragma1; pragma2; module; binding`.

Perhaps @alanz has an idea?

For now I've added module headers to the tests that were failing. I would like to add a known-failing test demonstrating the problem, but I don't know how to do that in the test framework.

michaelpj:
Otherwise this is ready for review @jaspervdj

michaelpj:
Probable GHC issue: https://gitlab.haskell.org/ghc/ghc/-/issues/23984

alanz:
I just added a test with

```
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE ScopedTypeVariables #-}
x = 1
```
To the ghc-exactprint  ghc-9.8 branch, and it passes.

The original example with `P1` and `P2` fails, because the pragma-extraction step errors with unknown pragmas.

I will be preparing a ghc-9.8.1 release this evening.

michaelpj:
Alan, I'm confused about how this ends up working for `stylish-haskell`. It depends on `ghc-lib-parser`, which is based on the GHC sources. I don't think it uses `ghc-exactprint` at all...

alanz:
@michaelpj I have no idea how stylish haskell works.  But if it does not use ghc-exactprint, then it may have to do something like https://github.com/alanz/ghc-exactprint/blob/master/src/Language/Haskell/GHC/ExactPrint/Parsers.hs#L303

michaelpj:
ping @jaspervdj ?

jaspervdj (approved):
Changes look good to me.  I would like to fix the module header issue, but I can do that separately as a follow-up if you want a release of this soon (e.g. for HLS)?

jaspervdj:
After reading through the comments and the GHC ticket more carefully, it's clear to me that fixing this downstream is probably more trouble than it's worth.  I will add one test case using an empty line rather than `module Main where` since I think that's more common in short `.hs` files.

michaelpj:
Thanks!

Would it be possible to get a release too?

jaspervdj:
Released as `v0.14.6.0`.  Thanks again for the GHC 9.8 PR!

PR #482, opened 2024-06-27
Make it possible to search for config without getCurrentDirectory

jhrcek:
An attempt to resolve https://github.com/haskell/stylish-haskell/issues/478

The intention is to make it possible to avoid dependency on getCurrentDirectory (which is problematic in haskell-language-server where multiple things running concurrently might depend on current directory and having to setCurrentDirectory just for stylish-haskell to be able to find the right config might lead to various race conditions).

The way to achieve it is to introduce new datatype `ConfigSearchStrategy` whose constructor `SearchFromDirectory startDir` makes it possible to supply a directory explicitly.

See this HLS PR for an example of how this new functionality would be used: https://github.com/haskell/haskell-language-server/pull/4338

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+- UNRELEASED
+     * #482 Add ConfigSearchStrategy to allow avoiding getCurrentDir when loading config (by Jan Hrček)
+
 - 0.14.6.0 (2024-01-19)
      * #471 Support GHC 9.8 (by Michael Peyton Jones)
      * #440 Fix dissappearing `DEPRECATED` pragma on module (by Lev Dvorkin)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,7 @@
 # CHANGELOG
 
 - UNRELEASED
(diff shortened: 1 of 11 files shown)


CHANGELOG
```diff
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+- UNRELEASED
+     * #482 Add ConfigSearchStrategy to allow avoiding getCurrentDir when loading config (by Jan Hrček)
```
jhrcek, line 4:
```suggestion
     * #482 Add ConfigSearchStrategy to allow avoiding getCurrentDirectory when loading config (by Jan Hrček)
```

jhrcek:
Hello @jaspervdj 
could you please check this when you have some time?

jaspervdj (approved):

jhrcek:
Converting to draft while I'm groping in the dark, trying to make CI pass..

jaspervdj:
Actually feel free to leave the CI changes out, I can try to get them in as a separate PR.

jhrcek:
Ok, thank you. 
Forgive my sausage fingers, I wanted to force push (-f) the branch with CI change commits reset but instead I deleted the branch in my remote (-d) :sweat_smile: I reopened the PR in https://github.com/haskell/stylish-haskell/pull/483

PR #483, opened 2024-06-30
Make it possible to search for config without getCurrentDirectory

jhrcek:
Reopening https://github.com/haskell/stylish-haskell/pull/482 whose branch I deleted by mistake.

Original description:
An attempt to resolve https://github.com/haskell/stylish-haskell/issues/478

The intention is to make it possible to avoid dependency on getCurrentDirectory (which is problematic in haskell-language-server where multiple things running concurrently might depend on current directory and having to setCurrentDirectory just for stylish-haskell to be able to find the right config might lead to various race conditions).

The way to achieve it is to introduce new datatype `ConfigSearchStrategy` whose constructor `SearchFromDirectory startDir` makes it possible to supply a directory explicitly.

See this HLS PR for an example of how this new functionality would be used: https://github.com/haskell/haskell-language-server/pull/4338


jaspervdj:
This looks good to me: but I think we should either

1.  Use something like `formatWith defaultFormatOptions {formatOptionsSearchStrategy = ...}`, and keep `format` as-is to not break backwards compatibility.

2.  Add an explicit example to the `CHANGELOG` of how code calling `format` should be ported.

I don't think many tools are calling `stylish-haskell` as a library, it's mostly used as an executable, so I think (2) is fine and (1) is probably overkill.

jhrcek:
I went for option 2 as it was easier. But do let me know if you change your mind and I can implement 1 instead. Intuitively I dislike 1 because (if I understand it right) we'd have to introduce a new type of config type that would have to be called differently than the original Config and which would (for now) just specify the config loading strategy?

Maybe this middle-ground solution would be enough to minimize backward api breakage?

```haskell
-- keep the original api
format :: Maybe ConfigPath -> Maybe FilePath -> String -> IO (Either String Lines)
format Nothing = formatWith SearchFromCurrentDirectory
format (Just cfgPath) = formatWith (UseConfig (unConfigPath cfgPath))

-- introduce new more general function
formatWith :: ConfigSearchStrategy -> Maybe FilePath -> String -> IO (Either String Lines)
formatWith configSearchStrategy maybeFilePath contents = do
  conf <- loadConfig (makeVerbose True) configSearchStrategy
  pure $ runSteps (configLanguageExtensions conf) maybeFilePath (configSteps conf) $ lines contents
```

fendor:
@jaspervdj friendly ping :)

jaspervdj:
Thanks for the ping and of course the PR!  This looks good to me.

jaspervdj:
Depends on #489

jaspervdj:
@jhrcek I've added one more case in 39a8c54.  It's good to merge now from my side if you're happy with that additional commit.

jhrcek:
Thank you, it looks great. I didn't realize cabal files were also searched in this way.

PR #491, opened 2025-04-06
Upgrade to ghc 9.12

GuillaumedeVolpiano:
Building on #480, this makes stylish-haskell functionally complete for ghc-9.12. Only one failing test left.

PR #493, opened 2025-04-14
Re-enabling Cabal-3.12

GuillaumedeVolpiano:
@jaspervdj : the bump to Cabal-3.14 creates an issue downstream with haskell-language-server, where many ghc-9.10 plugins are hard limited to Cabal-3.12. This patch attempts to adress the situation.

@jhrcek : I think this respects the logic behind your changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest]
-        ghc: ["9.6.6", "9.8", "9.10"]
+        ghc: ["9.6.6", "9.8", "9.10"] #using 9.6.6 as ghc-lib-parser-9.10.1.20250103 is not compatible with ghc-9.6.7. Re-enable when sorted
 
     steps:
     - uses: actions/checkout@v4
--- a/lib/Language/Haskell/Stylish/Config/Cabal.hs
+++ b/lib/Language/Haskell/Stylish/Config/Cabal.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 --------------------------------------------------------------------------------
 module Language.Haskell.Stylish.Config.Cabal
     ( findLanguageExtensions
@@ -50,11 +51,20 @@ findCabalFile verbose configSearchStrategy = case configSearchStrategy of
     verbose $ "Stylish Haskell will work basing on LANGUAGE pragmas in source files."
     return Nothing
   go searched (p : ps) = do
-    let projectRoot = Just $ Cabal.makeSymbolicPath p
+
+#if MIN_VERSION_Cabal(3,14,0)
+    let projectRoot = Just $ makeSymbolicPath p
     potentialCabalFile <- Cabal.findPackageDesc projectRoot
+#else
+    potentialCabalFile <- Cabal.findPackageDesc p
+#endif
     case potentialCabalFile of
       Right cabalFile -> pure $ Just $
+#if MIN_VERSION_Cabal(3,14,0)
         Cabal.interpretSymbolicPath projectRoot cabalFile
+#else
+        cabalFile
+#endif
       _ -> go (p : searched) ps
 
 
--- a/src/Main.hs
+++ b/src/Main.hs
@@ -1,9 +1,9 @@
 {-# LANGUAGE CPP #-}
+
 --------------------------------------------------------------------------------
 module Main
-    ( main
-    ) where
-
+  ( main
+  ) where
 
 --------------------------------------------------------------------------------
 import           Control.Monad            (forM_, unless, when)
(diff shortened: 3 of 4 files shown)


lib/Language/Haskell/Stylish/Config/Cabal.hs
```diff
@@ -50,11 +51,22 @@ findCabalFile verbose configSearchStrategy = case configSearchStrategy of
...
     return Nothing
   go searched (p : ps) = do
-    let projectRoot = Just $ Cabal.makeSymbolicPath p
+
+#if MIN_VERSION_Cabal(3,14,0)
+    let projectRoot = Just $ makeSymbolicPath p
     potentialCabalFile <- Cabal.findPackageDesc projectRoot
+#else
+    let projectRoot = Just $ Cabal.unsafeMakeSymbolicPath p
+        pkgDir = maybe "." Cabal.getSymbolicPath projectRoot
```
jhrcek, line 60: This `maybe "."` bit seems unnecessary. You know statically that projectRoot is Just, so why not call Cabal.getSymbolicPath directly on `Caba.lunsafeMakeSymbolicPath p`?

jhrcek (approved):
Extending version range of cabal dependency makes sense to me.
Could you please make sure hls can be built with these looser constraints (e.g. using source-repository-package referring to this PR's commit in hls repo similar to how [this PR](https://github.com/haskell/haskell-language-server/pull/4338/files#diff-b8ed06757045fac949c89f2139a862498ad2b6d1f82c61a31e7c91d6cf0eaa70R49-R52) does it?

lib/Language/Haskell/Stylish/Config/Cabal.hs
GuillaumedeVolpiano: Because I was being, let's say, litteral in my attempt to backport the logic of the your modifications to the Cabal-3.12 API.
Expanding everything gives

projectRoot = Just . SymbolicPath $ p

so

pkgdir = maybe "." (\(SymbolicPath path) -> path) $ Just . SymbolicPath $ p

Which boils down to

pkgdir = p. So I just went ahead and shunted the whole projectRoot part.

Tests downstream have started.

GuillaumedeVolpiano:
> Extending version range of cabal dependency makes sense to me. Could you please make sure hls can be built with these looser constraints (e.g. using source-repository-package referring to this PR's commit in hls repo similar to how [this PR](https://github.com/haskell/haskell-language-server/pull/4338/files#diff-b8ed06757045fac949c89f2139a862498ad2b6d1f82c61a31e7c91d6cf0eaa70R49-R52) does it?

Note that the tests will only work on ghc-9.10 because we have another issue : there is no hlint version compatible with ghc-lib-parser-9.10 (they went directly from 9.8 to 9.12)

lib/Language/Haskell/Stylish/Config/Cabal.hs
jhrcek: Just FYI, I wasn't involved in bumping lower bound of cabal - it seems it's been done by Jasper in https://github.com/haskell/stylish-haskell/commit/cdff43e8d85c4420954fccc0ee42ff2d600f806c#diff-f19a0f6fd6a7b06582842bce03f2ee66e3677c892e40c7d3433988b2f75bbb03R20-L41

GuillaumedeVolpiano: Yes sorry I keep confusing your commit and his

.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest]
-        ghc: ["9.6.6", "9.8", "9.10"]
```
jhrcek, line 12: Why are you removing the CI configs? stylish-haskell should probably still be compilabe with 9.6 and 9.8, right? It's just that hls with these changes no longer works with 9.6 and 9.8, right? So if we want new hls compatible version of stylish-haskell released, there's probably no need to remove these configurations from here..

jhrcek (approved):
Great. @jaspervdj could you please release 0.15.0.1 with these changes? We seem to need these relaxed version bound to be able to enable stylish in hls for ghc 9.10

GuillaumedeVolpiano:
The changes are also working for Cabal 3.10, so I relaxed the testing conditions, this way we still support ghc-9.6 and ghc-9.8

jaspervdj (approved):

.github/workflows/ci.yml
GuillaumedeVolpiano: I've re-added them.

GuillaumedeVolpiano:
So the issue with ghc-9.6.7 is with ghc-lib-parser. Reported here https://github.com/digital-asset/ghc-lib/issues/595
What do you want to do?

GuillaumedeVolpiano:
For what it's worth, there is the same issue with ghc-lib-parser-9.12.2 and ghc-9.8.4. The folks over at #ghc said to downgrade to ghc-lib-parser-9.8.5 (for ghc-9.6.7). I'm going to disable ghc-9.6 from the tests for now. If we want to include ghc 9.6 support, we are going to need to wait for a potential update of ghc-lib-parser-9.10

GuillaumedeVolpiano:
Well, I'll follow @jaspervdj and your lead on when is the best time to release. I've followed your advice and re-enabled 9.6.6 in the ci, with a note. Tests passed on my fork.

My 2 pence: we don't know if and when a new version of ghc-lib-parser-9.10 will be released, and there is no other reason the tests should fail on ghc-9.6.7. Also, we need this Cabal backport if we want to be able to use the ghc-lib-parser-9.12 version with <=ghc-9.10 (where we'll encounter the same issue with ghc-9.8.4 vs 9.8.3 for now, I think). As this is a fix, I think we could release, but once again, not my call on either side of the stream ;)

jaspervdj:
Thanks!  I am pushing one commit here to minimize the diff.  Then I'll release these changes, and after that I'll work on releasing #491.

jaspervdj:
Released in `v0.15.0.1` #494

